PR #3, opened 2020-05-25
[LOGGER] basic bare-bones logger completed

benliao1:
Logger is NOT a separate process; it's just an interface that processes can attach to and log stuff with. Consult `README.md` in `/logger` as well as the testers to see the usage!

Also fixed a few typos in shared memory wrapper and deleted / renamed some things, but didn't touch `dev_handler` or `executor` folder.


benliao1:
Yes, that makes sense. I left it as a TODO in the logger, but I suppose I could fix it now. I'll do that tonight, thanks!

benliao1:
Ok, I made the change and put some common things into the `runtime_util` folder. I also ran tests on both the logger and the wrapper to make sure they were still working.

PR #8, opened 2020-05-31
[SHM_WRAPPER] added shm_wrapper_aux and associated tests; bug fixes on device shm wrapper

benliao1:
- Added shm_wrapper_aux and suite of tests, including printing utilities; added Makefile
- Bug fixes on the synchronization function used to execute the shm wrapper tests consistently
- Added read and write functions for the device shm wrapper that allow user to reference devices with their 64-bit unique ID number, instead of their device index
- Added a file that just spawns the maximum number of devices in their shared memory wrapper for the executor to read
- Inserted logger messages into both shared memory wrappers

### DIFF
--- a/shm_wrapper/shm_wrapper.c
+++ b/shm_wrapper/shm_wrapper.c
@@ -297,6 +297,10 @@ void shm_init (process_t process)
 			error("sem_open: pmap_mutex@client");
 		}
 		
+		//wait on catalog_sem to ensure shm has been created before opening
+		if (sem_wait(catalog_sem) == -1) {
+			error("sem_wait: catalog_mutex@client");
+		}
 		
 		//open shared memory block and map to client process virtual memory
 		if ((fd_shm = shm_open(SHARED_MEM_NAME, O_RDWR, 0)) == -1) { //no O_CREAT
@@ -319,6 +323,11 @@ void shm_init (process_t process)
 			if ((sems[i].command_sem = sem_open((const char *) sname, 0, 0, 0)) == SEM_FAILED) { //no O_CREAT
 				error("sem_open: command sem@client");
 			}
+		}
+		
+		//release catalog_sem
+		if (sem_post(catalog_sem) == -1) {
+			error("sem_post: catalog_mutex@client");
 		}
 	}
 }
--- a/shm_wrapper/shm_wrapper.c
+++ b/shm_wrapper/shm_wrapper.c
@@ -50,6 +50,7 @@ sem_t *pmap_sem;                    //semaphore used as a mutex on the param bit
 static void error (char *msg)
 {
 	perror(msg);
+	log_runtime(ERROR, msg); //send the message to the logger
 	exit(1);
 }
 
@@ -104,7 +105,7 @@ void print_dev_ids ()
 	} else {
 		for (int i = 0; i < MAX_DEVICES; i++) {
 			if (catalog & (1 << i)) {
-				printf("dev_ix = %d: type = %d, year = %d, uid = %lu\n", i, dev_ids[i].type, dev_ids[i].year, dev_ids[i].uid);
+				printf("dev_ix = %d: type = %d, year = %d, uid = %llu\n", i, dev_ids[i].type, dev_ids[i].year, dev_ids[i].uid);
 			}
 		}
 	}
@@ -311,7 +312,7 @@ void device_connect (uint16_t dev_type, uint8_t dev_year, uint64_t dev_uid, int 
 		}
 	}
 	if (*dev_ix == MAX_DEVICES) {
-		printf("too many devices, connection unsuccessful\n"); //turn into a logger message later
+		log_runtime(WARN, "too many devices, connection unsuccessful");
 		if (sem_post(catalog_sem) == -1) { //release the catalog semaphore
 			error("sem_post: catalog_sem");
 		}
@@ -337,8 +338,8 @@ void device_connect (uint16_t dev_type, uint8_t dev_year, uint64_t dev_uid, int 
 
 	//reset param values to 0
 	for (int i = 0; i < MAX_PARAMS; i++) {
-		shm_ptr->params[0][*dev_ix][i] = (const param_val_t) {0};
-		shm_ptr->params[1][*dev_ix][i] = (const param_val_t) {0};
+		shm_ptr->params[DATA][*dev_ix][i] = (const param_val_t) {0};
+		shm_ptr->params[COMMAND][*dev_ix][i] = (const param_val_t) {0};
 	}
 	
 	//release associated data and command sems
@@ -375,19 +376,20 @@ void device_disconnect (int dev_ix)
 	if (sem_wait(sems[dev_ix].command_sem) == -1) {
 		error("sem_wait: command sem");
 	}
-
-
-	//update the catalog
-	shm_ptr->catalog &= (~(1 << dev_ix));
-
+	
+	//wait on pmap_sem
 	if (sem_wait(pmap_sem) == -1) {
 		error("sem_wait: pmap_sem");
 	}
 
-	//reset param map values to 0
-	shm_ptr->pmap[0] &= (~(1 << dev_ix)); //Reset the changed bit flag in pmap[0]
-	shm_ptr->pmap[dev_ix + 1] = 0; //turn off all changed bits for the device
+	//update the catalog
+	shm_ptr->catalog &= (~(1 << dev_ix));
 
+	//reset param map values to 0
+	shm_ptr->pmap[0] &= (~(1 << dev_ix)); //reset the changed bit flag in pmap[0]
+	shm_ptr->pmap[dev_ix + 1] = 0; //turn off all changed bits for the device
+	
+	//release pmap_sem
 	if (sem_post(pmap_sem) == -1) {
 		error("sem_post: pmap_sem");
 	}
@@ -422,10 +424,26 @@ No return value.
 */
 void device_read (int dev_ix, process_t process, stream_t stream, uint32_t params_to_read, param_val_t *params)
 {
+	char msg[64]; //for holding error message
+	
+	//wait on catalog_sem (TODO: check to see how much this affects performance with many threads)
+	if (sem_wait(catalog_sem) == -1) {
+		error("sem_wait: catalog_sem");
+	}
+	
 	//check catalog to see if dev_ix is valid, if not then return immediately
 	if (!(shm_ptr->catalog & (1 << dev_ix))) {
-		printf("no device at this index, read failed\n");
+		sprintf(msg, "no device at dev_ix = %d, read failed", dev_ix);
+		log_runtime(DEBUG, msg);
+		if (sem_post(catalog_sem) == -1) { //release the catalog_sem if no device
+				error("sem_post: catalog_sem");
+		}
 		return;
+	}
+	
+	//release catalog_sem
+	if (sem_post(catalog_sem) == -1) {
+		error("sem_post: catalog_sem");
 	}
 	
 	//grab semaphore for the appropriate stream and device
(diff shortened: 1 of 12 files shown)


shm_wrapper/shm_wrapper.c
```diff
@@ -311,7 +409,7 @@ void device_connect (uint16_t dev_type, uint8_t dev_year, uint64_t dev_uid, int
 		}
 	}
 	if (*dev_ix == MAX_DEVICES) {
-		printf("too many devices, connection unsuccessful\n"); //turn into a logger message later
+		log_runtime(WARN, "too many devices, connection unsuccessful");
```
AVSurfer123, line 412: Change this to call error. Shouldn't be a warning I think

shm_wrapper/shm_wrapper.c
```diff
@@ -422,57 +521,45 @@ No return value.
...
+	char msg[64]; //for holding error message
+	int dev_ix = -1;
 	
-	//read all requested params
-	for (int i = 0; i < MAX_PARAMS; i++) {
-		if (params_to_read & (1 << i)) {
-			params[i] = shm_ptr->params[stream][dev_ix][i];
+	//check catalog and shm_ptr->dev_ids to determine if device exists
+	for (int i = 0; i < MAX_DEVICES; i++) {
+		if ((shm_ptr->catalog & (1 << i)) && (shm_ptr->dev_ids[i].uid == dev_uid)) {
```
AVSurfer123, line 548: Would it be beneficial to have a map of uid -> idx because the executor will only be using these functions and I imagine the same for device handler?

shm_wrapper_aux/shm_wrapper_aux.c
```diff
@@ -0,0 +1,469 @@
...
+		rd_ptr->fields[DAWN] = DISCONNECTED;
+		rd_ptr->fields[SHEPHERD] = DISCONNECTED;
+		rd_ptr->fields[GAMEPAD] = DISCONNECTED;
+		rd_ptr->fields[TEAM] = BLUE;                //arbitrary
+		rd_ptr->net_to_exe_pending = 0;             //no writes made yet
+		rd_ptr->exe_to_net_pending = 0;             //no writes made yet
+		
+		//initialization complete; set gp_mutex to 1 indicating shm segment available for client(s)
+		if (sem_post(gp_sem) == -1) {
+			error("sem_post: gamepad_mutex@net_handler");
```
AVSurfer123, line 212: 1. Why is this needed? 2. Why don't you also post rd_sem?

shm_wrapper_aux/shm_wrapper_aux.c
```diff
@@ -0,0 +1,469 @@
...
+		if (process == NET_HANDLER && !(rd_ptr->net_to_exe_pending)) {
+			break;
+		} else if (process == EXECUTOR && !(rd_ptr->exe_to_net_pending)) {
+			break;
+		}
+		usleep(1000);
+		gettimeofday(&curr, NULL);
+		time_taken = (double)(curr.tv_sec - start.tv_sec) + ((double)(curr.tv_usec - start.tv_usec) / 1000000.0);
+	}
+	if (time_taken >= DESC_WR_MAXBLK) {
```
AVSurfer123, line 358: I don't think it's a good idea to stall like this. I don't see any use case where this is needed, and it seems to hurt in most cases.

shm_wrapper_aux/shm_wrapper_aux.c
```diff
@@ -0,0 +1,469 @@
...
+void gamepad_read (process_t process, uint32_t *pressed_buttons, float *joystick_vals)
+{
+	//wait on rd_sem
+	if (sem_wait(rd_sem) == -1) {
+		error("sem_wait: robot_desc_mutex");
+	}
+	
+	//if no gamepad connected, then release rd_sem and return
+	if (rd_ptr->fields[GAMEPAD] == DISCONNECTED) {
+		log_runtime(WARN, "tried to read, but no gamepad connected");
```
AVSurfer123, line 398: Think this should be an error.

shm_wrapper_aux/shm_wrapper_aux.c
```diff
@@ -0,0 +1,469 @@
...
+void gamepad_write (process_t process, uint32_t pressed_buttons, float *joystick_vals)
+{
+	//wait on rd_sem
+	if (sem_wait(rd_sem) == -1) {
+		error("sem_wait: robot_desc_mutex");
+	}
+	
+	//if no gamepad connected, then release rd_sem and return
+	if (rd_ptr->fields[GAMEPAD] == DISCONNECTED) {
+		log_runtime(WARN, "tried to write, but no gamepad connected");
```
AVSurfer123, line 443: Like above, think this should be an error.

shm_wrapper_aux/shm_wrapper_aux.h
```diff
@@ -0,0 +1,112 @@
...
+} gp_button_t;
+
+//enumerated names for the joystick params of the gamepad
+typedef enum gp_joysticks {
+	X_LEFT_JOYSTICK, Y_LEFT_JOYSTICK, X_RIGHT_JOYSTICK, Y_RIGHT_JOYSTICK
+} gp_joystick_t;
+
+//enumerated names for the different values the robot description fields can take on
+typedef enum robot_desc_vals {
+	ISSUE, NOMINAL,             //values for robot.state
```
AVSurfer123, line 33: What is the purpose of this field?

shm_wrapper_aux/shm_wrapper_aux.h
```diff
@@ -0,0 +1,112 @@
...
+#include <semaphore.h>                     //for semaphores
+#include <sys/mman.h>                      //for posix shared memory
+#include "../logger/logger.h"              //for logger (TODO: consider removing relative pathname in include)
+#include "../runtime_util/runtime_util.h"  //for runtime constants (TODO: consider removing relative pathname in include)
+
+#define NUM_DESC_FIELDS 6                   //number of fields in the robot description
+#define NUM_GAMEPAD_BUTTONS 17              //number of gamepad buttons
+
+//enumerated names for the buttons on the gamepad
+typedef enum gp_buttons {
```
AVSurfer123, line 21: Put these 4 enums into runtime util

shm_wrapper_aux/shm_wrapper_aux.h
```diff
@@ -0,0 +1,112 @@
...
+
+//enumerated names for the joystick params of the gamepad
+typedef enum gp_joysticks {
+	X_LEFT_JOYSTICK, Y_LEFT_JOYSTICK, X_RIGHT_JOYSTICK, Y_RIGHT_JOYSTICK
+} gp_joystick_t;
+
+//enumerated names for the different values the robot description fields can take on
+typedef enum robot_desc_vals {
+	ISSUE, NOMINAL,             //values for robot.state
+	IDLE, AUTO, TELEOP,         //values for robot.run_mode
```
AVSurfer123, line 34: Could you also add ESTOP?

benliao1: I figured that ESTOP could be implemented by just having the network handler set this field to IDLE, because that's what ESTOP really does. So rather than have you have to deal with doing the same thing for either this thing being in IDLE or ESTOP you just make sure that the student code isn't running when idling.

shm_wrapper_aux/shm_wrapper_aux.h
benliao1: I thought this might be useful for the network handler to know when it needs to send logger outputs, like maybe we could have some default option that's like only send logger output when the robot state goes to ISSUE and otherwise don't send logger output at all? Idk I added it for potential convenience, but if it doesn't end up getting used then I guess it's not necessary.

shm_wrapper/shm_wrapper.c
benliao1: We could, but like it would require building / maintaining / searching an array each time, whereas this is just looping through an array so I feel like this is the faster solution? Unless I'm misunderstanding what you're saying.....

shm_wrapper/shm_wrapper.c
benliao1: Makes sense, will change~~

shm_wrapper_aux/shm_wrapper_aux.c
benliao1: So it only stalls when the previously written value hasn't been read by the receiving end yet, i.e. under normal circumstances it won't stall at all. The use case is like maybe there's some command to go into a particular state or perform some action that is sent to the robot from Dawn or Shepherd, and it gets written to that particular field (some of these might be like game-specific things, like my freshman year the game had something like this where the robot had to do a certain thing once when Shepherd gave them the command to do it). But before that command gets read by the executor, the network handler writes to that block again, overwriting the previous value, and so the robot just misses the first value completely and doesn't execute the command. It's just a safety measure to try and ensure that all the things written to the robot description block are passed on to the desired receiver. But if everything is functioning normally, it doesn't stall at all.

shm_wrapper_aux/shm_wrapper_aux.c
benliao1: 1) It's needed because I initialize gp_sem with an initial value of 0 on line 159. Basically it makes sure that all the shared memory is created and initialized before other processes try to attach to it, because other processes will wait to open the gp_sem and that will block onto gp_sem is posted.
2) I don't need to post rd_sem because I initialize it with an initial value of 1 (unlocked) on line 163. I arbitrarily chose gp_sem to be the one that blocks other processes from trying to attach to shared memory before it's created.

shm_wrapper/shm_wrapper.c
AVSurfer123: I meant like a hashmap from device uid -> device idx. Since the UIDS are random, should be pretty spread hash buckets. But that might be a lot of work in C and the array size is fixed at 32 anyways. I think this is fine.

shm_wrapper_aux/shm_wrapper_aux.c
AVSurfer123: In almost all cases though, we don't want to read every update. Like we don't care if the gamepad went from connected to disconnected to connected. same for dawn and shepherd states. The existing states are not important enough to read every single write from what I've seen and the network handler shouldn't be writing states extremely quickly either. Even if there was a use case I'm missing, the block is written incorrectly because it blocks every single write to all fields when it should only block the write to the specific field that changed. And same for the read flag.

shm_wrapper_aux/shm_wrapper_aux.h
AVSurfer123: We always need to send logger output because the print statements the student uses gets sent to the logger.

shm_wrapper_aux/shm_wrapper_aux.h
AVSurfer123: ok thats fine I guess. I was just wondering if there was ever a case where we need to turn the runtime processes off, which would be ESTOP, but hopefully not if our code is robust lol.

shm_wrapper_aux/shm_wrapper_aux.c
benliao1: Done

shm_wrapper_aux/shm_wrapper_aux.c
benliao1: Done

shm_wrapper/shm_wrapper.c
benliao1: Done

shm_wrapper_aux/shm_wrapper_aux.h
benliao1: Done

shm_wrapper/shm_wrapper.c
benliao1: ok sounds good, I mean this could be something we investigate later, it might be possible to get some speedup with this. But for now I think it's fine, like you said~~

shm_wrapper_aux/shm_wrapper_aux.c
benliao1: Ok I see what you're saying, like they shouldn't be writing something once and then reverting it back very quickly. I'll change this...

shm_wrapper_aux/shm_wrapper_aux.h
benliao1: Ok, good point. I wrote the wrapper so that it's easy to add new fields as we need, so I'll delete them for now since we can't see a use case for it at the moment!

PR #9, opened 2020-06-04
[SHM_WRAPPER] minor cleanups and additions to shm wrappers

benliao1:
1) Added `STUDENTAPI` as a possible process
2) Cleaned up semaphore logic in both wrapper significantly (considering putting the `my_sem_*` functions in `runtime_util` actually, but for now they can stay in the wrappers)
3) Removed `process` argument from `gamepad_read` and `gamepad_write` functions

### DIFF
--- a/logger/logger.c
+++ b/logger/logger.c
@@ -38,7 +38,6 @@ void logger_init (process_t process)
 	} else if (process == NET_HANDLER) {
 		sprintf(process_str, "NET_HANDLER");
 	}
-	
 }
 
 void logger_stop ()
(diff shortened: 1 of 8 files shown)


logger/logger.c
```diff
@@ -37,8 +37,9 @@ void logger_init (process_t process)
 		sprintf(process_str, "EXECUTOR");
 	} else if (process == NET_HANDLER) {
 		sprintf(process_str, "NET_HANDLER");
+	} else if (process == STUDENTAPI) {
```
AVSurfer123, line 40: can you change it to `API` cus I'm too lazy to change my code

benliao1: We should try to be clear on what API we're talking about I think :P

PR #13, opened 2020-06-13
Executor Completed, Version 1.0

AVSurfer123:
The code executor has been completed. The `executor.c` file will wait for mode changes from the network handler and will run the corresponding mode in a separate thread by calling the main and setup functions in `studentcode.py`. These functions will call the student API which is written in Cython in `studentapi.pyx` which will talk to shared memory. 

Core files:
executor.c
studentapi.pyx
studentapi.pxd

Build files:
Makefile
setup.py

Test files:
studentcode.py
test_studentapi.py
../create_static_shm.sh


executor/studentapi.pyx
```diff
@@ -0,0 +1,238 @@
...
+        log_runtime(DEBUG, "SHM intialized")
+        self.running_actions = {}
+
+    def __dealloc__(self):
+        """Once process is done and object is deallocated, close the mapping to SHM."""
+        shm_stop(STUDENTAPI)
+        log_runtime(DEBUG, "SHM stopped")
+
+
+    def run(self, action, *args, **kwargs) -> None:
```
jonathan-j-lee, line 146: FYI, some subtleties with actions:
1. [`Robot.run`](https://pioneers.berkeley.edu/software/robot_api.html#robot-run-async-function-args-) is supposed to accept a coroutine function (defined with `async def`), so accepting regular functions for `action` would break existing student code. Threads are not a drop-in replacement for [awaitables](https://docs.python.org/3/library/asyncio-task.html#awaitables).
1. There's no guarantee a student-supplied function will terminate quickly, if at all. In fact, the rationale for actions is to support [long-running time-sensitive tasks](https://pioneers.berkeley.edu/software/auto.html) outside of the `setup/main` loop. Counting on actions to terminate eventually is not a good solution, IMO.
1. For device contention reasons, an action should only have one copy running at a [given time](https://github.com/pioneers/PieCentral/blob/master/runtime/runtime/studentapi.py#L206).

The original action executor implementation was carefully designed to avoid these problems. A dedicated thread runs student actions using `asyncio`, and the actions are scheduled from the main thread executing `setup/main`. Unlike threads, awaitables can be safely cancelled.

executor/code_loader.py
```diff
@@ -0,0 +1,112 @@
...
+import typing
+import studentcode
+
+
+def sigalarm_handler(signum, frame):
+    """Handler function that is called when process is interrupted by a SIGALRM. Causes process to cleanly exit."""
+    raise TimeoutError
+
+# To register the handler for the specified signal
+signal.signal(signal.SIGALRM, sigalarm_handler)
```
jonathan-j-lee, line 16: This will set a signal handler even if you just import the module. We should avoid hidden side effects by moving the registration under the `__name__ == '__main__'` guard.

executor/code_loader.py
```diff
@@ -0,0 +1,112 @@
...
+        """Runs the student code with the predefined timeouts. First calls `mode_setup()` then calls `mode_main()` based on the `mode` attribute. The timeout is handled using SIGALRM signals."""
+        if self.mode != 'autonomous' and self.mode != 'teleop':
+            raise CodeExecutionError("Only modes that can be run are autonomous and teleop.")
+        setup = self.get_function_safe(f'{self.mode}_setup')
+        main = self.get_function_safe(f'{self.mode}_main')
+        start = time.time()
+        try:
+            self.run_with_timeout(setup, self.setup_timeout)
+        except Exception as e:
+            signal.setitimer(signal.ITIMER_REAL, 0)
```
jonathan-j-lee, line 80: `run_with_timeout` already clears the alarm.

executor/code_loader.py
```diff
@@ -0,0 +1,112 @@
+import time
+import importlib
+import sys
+import signal
+import studentapi
+from studentapi import CodeExecutionError, _print
+import typing
+import studentcode
```
jonathan-j-lee, line 8: What is the purpose of importing `studentcode` here if it's dynamically imported using `importlib`?

executor/code_loader.py
```diff
@@ -0,0 +1,112 @@
...
+        diff = time.time() - start
+
+        signal.setitimer(signal.ITIMER_REAL, self.TIMEOUT[self.mode] - diff)
+        try:
+            while True:
+                start = time.time()
+                main()
+                diff = time.time() - start
+                if diff > self.interval:
+                    raise CodeExecutionError(f"An iteration of {self.mode}_main took {diff} seconds, longer than {self.interval} seconds, indicating student code is stuck in a loop.")
```
jonathan-j-lee, line 91: Runtime should continue running student code to the best of its ability and log the poor performance instead of exiting early. This may cause robots to stop during a match due to the idiosyncrasies of the CPU scheduler.

executor/code_loader.py
```diff
@@ -0,0 +1,112 @@
...
+                while((time.time() - start) < self.interval):
+                    pass
```
jonathan-j-lee, line 93: Very bad to busy-wait; it gives the OS the impression this process is doing useful work. `time.sleep` instead.

executor/studentapi.pyx
```diff
@@ -0,0 +1,238 @@
...
+            value: Value to set for the param. The type of the value can be seen at https://pioneers.berkeley.edu/software/robot_api.html
+        """
+        param_bytes = param_name.encode('utf-8')
+        cdef char* param = param_bytes
+        splits = device_id.split('_')
+        cdef int device_type = int(splits[0])
+        cdef uint64_t device_uid = int(splits[1])
+        cdef param_desc_t* param_desc
+        cdef uint8_t param_idx
+        with nogil:
```
jonathan-j-lee, line 223: The purpose of `with nogil` is to allow Python multithreading to use multiple physical cores simultaneously instead of just multiplexing in time. The caveat is that the body must not interact with Python objects.

You'll probably actually incur a penalty for releasing, then reacquiring, the GIL for such short amounts of time very frequently. Unless you're heavily multithreading with CPU-intensive tasks, best to hold onto the GIL.

executor/code_loader.py
AVSurfer123: Hey, I should've deleted `code_loader.py` before, but this was just the initial version I tried load the code. I don't use this file anymore and it's now deprecated so I'm going to delete it now.  All the logic for calling the student code in threads is now in `executor.c` if you want to check that out. Same thing for the rest of the comments on `code_loader` so I'm going to resolve them.

executor/studentapi.pyx
AVSurfer123: So you think its better if I remove the GIL release? I was originally thinking it could help if the student had multiple actions running at the same time, but I guess its not enough of an increase. A while ago, I tried testing if adding it slowed the code down and if I remember it was slightly slower, maybe like 5-10%.

jonathan-j-lee: `nogil` is not necessary if actions are implemented with `asyncio` (which we must do anyway if we want backwards-compatbility). `asyncio` is nice because it's single-threaded, meaning GIL contention is not even relevant, since the OS does not handle concurrency. Using `asyncio` for actions, you'll only need two threads: the main thread for `setup`/`main` (synchronous, uses `SIGALRM` to interrupt long-running code), and an auxiliary daemon thread for actions.

executor/code_loader.py
jonathan-j-lee: OK, yeah, that's what I figured. I don't know enough about Python's C API to help.

executor/studentapi.pyx
AVSurfer123: 1. Yea I know this. I thought it would simplify the student code in the future though it is backward incompatible. I also think its better if the students don't have to worry about using `async` at all, if its possible.
2. Yea I agree. I'm thinking of now creating a flag/exception wrapper around the student's function so that I can cancel the actions safely without doing `pthread_cancel`. I have a similar issue with the main thread to safely cancel between auton and teleop that I still need to figure out.
3. I didn't think about this, but I can change it so that if the action is currently running it will throw an error/warning. What behavior do you think should happen when the student schedules multiple actions of the same name?

jonathan-j-lee: 1. I agree that `async`/`await` syntax has historically caused a lot of confusion for students and within PiE, but this can be overcome with better documentation. `asyncio` offers several [advantages](https://discuss.python.org/t/what-are-the-advantages-of-asyncio-over-threads/2112/6) over manual thread management for our use case (especially cancellation, see below).
1. Student code is completely opaque, so using `threading.Event` or `queue.Queue` to pass halt messages in the standard way is not possible without a massively complex API change. If we insist on synchronous actions, [a thread pool](https://docs.python.org/3/library/concurrent.futures.html) caps the number of workers, but even a pool cannot terminate tasks when going back to idle (see the note on `shutdown`). The executor process would need to exit and be respawned.
1. If the action is already running, simply log the attempt at `DEBUG`. It's not anomalous to schedule an action multiple times, since we allow `Robot.run` to be called from `(autonomous|teleop)_main`, which is called very frequently. `Robot.run`'s semantics follow "run, if not already running". [Staff Curry](https://github.com/pioneers/PieCentral/blob/master/staff-bots/staff_curry_code.py#L67-L84) illustrates this well: the button for waving might be pressed for many consecutive cycles, but obviously it's all part of one wave. The unfortunate reality is that very few people remember to use `Robot.is_running(action)`.

The original [API design notes](https://drive.google.com/file/d/0B0UMIFiWC3aJZnF1WVdfZTd3b1U/view?usp=sharing), in case you find them helpful.

AVSurfer123: 2. I think I figured a way to cancel the Python threads safely. It uses `PyThreadState_SetAsyncExc`, which is a Python C API function, to send an exception to a thread. You can see the code for this in the `ThreadWrapper` in `studentapi.pyx` and in `actions_handler` in `executor.c`. I do agree though that `asyncio` is probably safer and perhaps faster due to OS overhead. I might implement that later instead of `threading` if `threading` has issues, but at least in my testing, it seems to work.
3. Changed the function to these semantics.

executor/studentapi.pyx
AVSurfer123: I ended up removing the GIL release blocks. For now, I don't think I'll be using asyncio but you're right that the time spent to read from shared memory is so small that releasing and reacquiring the lock seems pointless.

executor/executor.c
```diff
@@ -0,0 +1,492 @@
...
+                char msg[128];
+                sprintf(msg, "Python function %s call failed", args->func_name);
+                log_runtime(ERROR, msg);
+                break;
+            }
+            else {
+                Py_DECREF(pValue);
+            }
+        } while(args->loop);
+        pthread_setcanceltype(PTHREAD_CANCEL_DEFERRED, NULL);
```
benliao1: Is there a reason why we don't set the cancel type back to `DEFERRED` earlier than this? We only want it to be asynchronous cancel when it's in the `PyObject_CallObject()` function right?

executor/executor.c
```diff
@@ -0,0 +1,492 @@
...
+
+        // Ensure we only cancel threads that are still alive
+        PyObject* alive = PyObject_CallMethod(thread, "is_alive", NULL);
+        if (alive == NULL) {
+            PyErr_Print();
+            log_runtime(ERROR, "Couldn't get whether thread is alive");
+            continue;
+        }
+        else if (PyObject_Not(alive)) {
+            continue;
```
benliao1: Is a `Py_DECREF(alive);` necessary here?

executor/studentapi.pyx
```diff
@@ -0,0 +1,272 @@
...
+    def __cinit__(self, mode):
+        """Initializes the mode of the robot. Also initializes the auxiliary SHM. """
+        self.mode = mode
+        # shm_aux_init(STUDENTAPI)
+        # log_runtime(DEBUG, "Aux SHM initialized")
+
+    def __dealloc__(self):
+        """Once process is finished and object is deallocated, close the mapping to the auxiliary SHM."""
+        # shm_aux_stop(STUDENTAPI)
+        # log_runtime(DEBUG, "Aux SHM stopped")
```
benliao1, line 101: So we straightup don't need this any more? Do we still need this function?

executor/studentapi.pyx
```diff
@@ -0,0 +1,272 @@
...
+        """Initializes the shared memory (SHM) wrapper. """
+        # shm_init(STUDENTAPI) # Remove once integrated into executor?
+        # log_runtime(DEBUG, "SHM intialized")
+        self.running_actions = {}
+
+    def __dealloc__(self):
+        """Once process is done and object is deallocated, close the mapping to SHM."""
+        # shm_stop(STUDENTAPI)
+        # log_runtime(DEBUG, "SHM stopped")
+
```
benliao1, line 163: Same comment here as with the Gamepad class~~

runtime_util/runtime_util.h
```diff
@@ -31,9 +32,10 @@ typedef enum gp_joysticks {
 	X_LEFT_JOYSTICK, Y_LEFT_JOYSTICK, X_RIGHT_JOYSTICK, Y_RIGHT_JOYSTICK
 } gp_joystick_t;
 
+
 //enumerated names for the different values the robot description fields can take on
 typedef enum robot_desc_vals {
-	IDLE, AUTO, TELEOP,         //values for robot.run_mode
+	IDLE, AUTO, TELEOP, ESTOP,       //values for robot.run_mode
```
benliao1, line 38: Did we end up needing the `ESTOP` value? I don't see it anywhere in your code so I'm assuming no? If we don't need it we should delete it~~

executor/executor.c
```diff
@@ -0,0 +1,491 @@
...
+
+		// Ensure we only cancel threads that are still alive
+		PyObject* alive = PyObject_CallMethod(thread, "is_alive", NULL);
+		if (alive == NULL) {
+			PyErr_Print();
+			log_runtime(ERROR, "Couldn't get whether thread is alive");
+			continue;
+		}
+		else if (PyObject_Not(alive)) {
+			continue;
```
benliao1, line 432: Do we need a `Py_DECREF(alive);` here before the `continue;`?

executor/executor.c
```diff
@@ -0,0 +1,491 @@
...
+				char msg[128];
+				sprintf(msg, "Python function %s call failed", args->func_name);
+				log_runtime(ERROR, msg);
+				break;
+			}
+			else {
+				Py_DECREF(pValue);
+			}
+		} while (args->loop);
+		pthread_setcanceltype(PTHREAD_CANCEL_DEFERRED, NULL);
```
benliao1, line 284: Is there a reason why we wait until here to set the cancel type back to `DEFERRED`? I thought we only want it asynchronous when the Python function is called, in `PyObject_CallObject()`... if that's the case, the two calls to `pthread_canceltype()` should probably go between lines 259 and 260, and between lines 260 and 261, to wrap the call to the Python function.

executor/executor.c
AVSurfer123: yea I forgot to decref stuff in that function. I'll fix it

executor/studentapi.pyx
AVSurfer123: No

runtime_util/runtime_util.h
AVSurfer123: Executor wouldn't be using it, only the supervisor would.

executor/executor.c
AVSurfer123: Fixed

executor/studentapi.pyx
AVSurfer123: Removed

executor/executor.c
AVSurfer123: Yea fixed.

executor/executor.c
AVSurfer123: Fixed

Issue #24, opened 2020-06-27
Script to print out current SHM and Aux SHM values


AVSurfer123:
This script would be very useful now to check that the net handler and dev handler are working.

benliao1:
Was I going to do this? Or @AVSurfer123 ? Also perhaps not necessary due to the `print_gamepad_state()` and `print_robot_desc` functions in Aux SHM, and the `print_catalog`, `print_params`, and `print_dev_ids` functions in SHM.

AVSurfer123:
No, I can do it. I just wanted an executable file that I can run to tell me it. Takes like 2 lines

Issue #22, opened 2020-06-27
Convert `param_val_t` from struct to union


AVSurfer123:
This should be done as we use the struct as a union, with only 1 of its members being used at a time. This should would impact `studentapi.pyx` the most, and might have slight modifications to `shm_wrapper.c`.

Issue #21, opened 2020-06-26
[LOGGER] deprecate log_runtime


benliao1:
Remove all calls to `log_runtime` in the codebase and deprecate it in `logger.c` / `logger.h`

benliao1:
Also add `strerror` support in `log_printf` to make it easier to output error-specific strings to log messages

Issue #19, opened 2020-06-24
[EXECUTOR] fix compilation warnings


benliao1:
Fix perhaps small code semantic errors that cause compilation warnings in `executor` and `connect_static_devices_aux`. Shouldn't take too long, but very low priority.

PR #32, opened 2020-06-29
Added utility print scripts + code cleanup

### LINKED ISSUE #23: [AUX_SHM] Remove all TEAM related code
Runtime doesn't need to know what team the robot is in. Need to remove the enum for it in `runtime_util.h` and also fix the corresponding code in `shm_wrapper_aux.c`.

AVSurfer123:
Fixes #23 #24 #19 #22 . Added print_shm and print_aux_shm to see the state of both shared memory blocks. Cleaned up their Makefiles and made them OS agnostic. Got rid of all `log_runtime` calls within `executor` to help with #21 .  Removed old files in shm_wrapper. Added error checking for logger calls to ensure it is initialized. Added `TEST` process type for testing


shm_wrapper/print_shm.c
```diff
@@ -0,0 +1,16 @@
...
+        devices = -1;
+    }
+    else {
+        for (int i = 1; i < argc; i++) {
+            devices |= (1 << atoi(argv[i]));
+        }
+    }
+    logger_init(TEST);
+    shm_init(TEST);
+    print_params(devices);
```
benliao1, line 15: should also `shm_stop(TEST)` and `logger_stop(TEST)` here just to be explicit... especially if logging to network (not sure how hanging open pipes will behave--probably it's ok because a process closes all file descriptors when it exits, but just to be safe)

executor/executor.c
```diff
@@ -146,10 +146,14 @@ void executor_init(char* student_code) {
...
         PyErr_Print();
-        log_runtime(ERROR, "Could not insert API into student code.");
+        log_printf(ERROR, "Could not insert API into student code.");
         executor_stop();
     }
 
+    sigset_t sigset;
+    sigemptyset(&sigset);
+    sigaddset(&sigset, SIGINT);
+    sigprocmask(SIG_BLOCK, &sigset, NULL);
```
benliao1, line 156: What are these for?

shm_wrapper_aux/print_aux_shm.c
```diff
@@ -0,0 +1,8 @@
+#include "shm_wrapper_aux.h"
+
+int main() {
+    logger_init(TEST);
+    shm_aux_init(TEST);
+    print_robot_desc();
+    print_gamepad_state();
+}
```
benliao1, line 8: Same comment as above here~~

shm_wrapper/print_shm.c
AVSurfer123: Yea forgot about them. I'll add them

PR #37, opened 2020-07-06
Network Handler 1.0 Completed

AVSurfer123:
Version 1.0 of the network handler is finished. It works by having separate TCP and UDP connections. There is 1 TCP and 1 UDP connection with Dawn and 1 TCP connection with Shepherd. Since UDP works with whole packets, the UDP thread is started when net_handler starts and dies with net_handler dies. The TCP threads rely on connections so whenever a client connects/disconnects, the corresponding TCP thread starts/stops. More detail on specifics can be found in the README.

Additional, 2 major system overhauls 
1. the logger which will also output to the network and has a new config system
2. the executor which will also handle running code challenges

Structure:
protos: contains the Protobuf definitions of all our network messages
pbc_gen: the corresponding C code that is generated from the Protobufs using protobuf-c
protobuf_tests: contains several test files to ensure that the Protobuf message passing/reading works
test: contains test clients for TCP and UDP that imitate Dawn and Shepherd

net_handler.c - main entry file that starts UDP and connects to TCP
net_util.c - helper functions to communicate with the TCP sockets
tcp_conn.c - handles the TCP connection
udp_conn.c - handles the UDP connection


benliao1:
I tested it decently thoroughly, the code looks pretty solid (more documentation / clarity could be added to UDP suite) so I'll approve it for now and then in my testing framework branch I'll work to resolve any issues that might come up, and document the UDP stuff a bit more!

net_handler/tcp_conn.c
```diff
@@ -181,11 +183,10 @@ static int recv_new_msg (int conn_fd, int challenge_fd)
 		}
 		text__free_unpacked(inputs, NULL);
 
-		log_printf(DEBUG, "entering CHALLENGE mode. running coding challenges!");
+		log_printf(INFO, "entering CHALLENGE mode. running coding challenges!");
```
AVSurfer123, line 186: I don't think any of these statements should be INFO. I feel like we should almost never send logs from runtime to the student unless its an error.

net_handler/tcp_conn.c
```diff
@@ -152,14 +149,19 @@ static int recv_new_msg (int conn_fd, int challenge_fd)
...
 	else if (err == -1) { // Means there is some other error while reading
 		return -2; 
 	}
+	
 	//unpack according to message
 	if (msg_type == CHALLENGE_DATA_MSG) {
+		//socket address structure for the UNIX socket to executor for challenge data
+		struct sockaddr_un exec_addr;
+		exec_addr.sun_family = AF_UNIX;
+		strcpy(exec_addr.sun_path, CHALLENGE_SOCKET);
```
AVSurfer123, line 163: I think you should use the struct initializer I had earlier, or use memset

net_handler/net_handler.c
```diff
@@ -156,18 +157,20 @@ int main ()
...
 			continue;
 		}
-		//if the incoming request is dawn or shepherd, start the appropriate threads
-		if (is_dawn && robot_desc_read(DAWN) == DISCONNECTED) {
-			log_printf(DEBUG, "Starting Dawn connection");
-			start_tcp_conn(DAWN, connfd, 1);
-		} else if (robot_desc_read(SHEPHERD) == DISCONNECTED) {
+		
+		//if the incoming request is shepherd or dawn, start the appropriate threads
+		if (client_id == 0 && robot_desc_read(SHEPHERD) == DISCONNECTED) {
```
AVSurfer123, line 168: Not sure if you want this, but I was thinking about also checking the port here as the ports are something that are easily binded by both sides. Adds a little bit of extra security. Either way though, we should delete the two functions `is_dawn` and `is_shepherd` from this file.

executor/executor.c
```diff
@@ -481,5 +481,6 @@ int main(int argc, char* argv[]) {
                 }
             }
         }
+		usleep(4000); //throttle this thread to ~200 loops/sec
```
AVSurfer123, line 484: Make it 100,000 i.e. 10 Hz. The main loop doesn't need to be that responsive

net_handler/tcp_conn.c
benliao1: I feel like it would be useful to know, since like suppose a student uploads some code and runs it and presses "Autonomous" on Dawn. If they get nothing on the console, the student might think that something is broken with runtime or the connection is bad; if we do send something (entering <mode> mode) and then it still prints nothing then they know that the system is working fine and it's their code that's breaking.

net_handler/tcp_conn.c
benliao1: This generated errors when I compiled it on my Mac (saying you can't initialize a`char *` with non-constant array or something like that). I looked it up online and it seems that it's good practice to do it this way (i.e. use `strcpy`. This way also makes it clear what the names of the fields that we're initializing are, as opposed to `struct sockaddr_un exec_addr = { AF_UNIX, CHALLENGE_SOCKET }` which doesn't tell you what the names of the fields are (`sun_family` and `sun_path`)

net_handler/net_handler.c
benliao1: 👌

executor/executor.c
benliao1: 👌

net_handler/net_handler.c
benliao1: Actually, I'm going to keep `is_dawn` and `is_shepherd` because we still have 3 conditions for starting the thread, it's just different from the original 3 conditions. Our new conditions are:
Is `client_id` valid?

net_handler/tcp_conn.c
AVSurfer123: you should still use memset to set rest of the elements to 0 I believe. I know thats good practice for sockaddr_in atleast.

net_handler/net_handler.c
AVSurfer123: Sure but theres no need for 2 extra functions that only get used in 1 place. Just makes the code unnecessarily complex

net_handler/tcp_conn.c
```diff
@@ -138,7 +144,7 @@ static void send_challenge_results(int conn_fd, int challenge_fd) {
  *    - int results_fd: file descriptor of FIFO pipe to executor to which to write challenge input data if received
  * Returns: pointer to integer in which return status will be stored
  *      0 if message received and processed
- *     -1 if message could not be parsed because Shepherd disconnected and connection closed
+ *     -1 if message could not be parsed because ckuebt disconnected and connection closed
```
AVSurfer123, line 147: typo? xD

AVSurfer123: Fixed it

net_handler/tcp_conn.c
AVSurfer123: Fixed it

net_handler/net_handler.c
AVSurfer123: Fixed it

PR #51, opened 2020-07-14
[LOGGER] Remove all log_runtime in project

benliao1:
Removed `log_runtime` from project and converted to `log_printf`. `log_runtime` is not in `logger` any more either.

### DIFF
--- a/logger/logger.c
+++ b/logger/logger.c
@@ -32,9 +32,8 @@ char log_file_path[MAX_CONFIG_LINE_LEN];    //file path to the log file
 
 //used for LOG_NETWORK only
 mode_t fifo_mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;   // -rw-rw-rw permission for FIFO
-int fifo_up = 0;                                         //flag that represents if FIFO is up and running
-int fifo_fd = -1;                                        //file descriptor for FIFO
-// char fifo_path[MAX_CONFIG_LINE_LEN] = "/tmp/log-fifo";   //file path to the FIFO pipe for logs
+int fifo_up = 0;      //flag that represents if FIFO is up and running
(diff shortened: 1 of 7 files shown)


logger/logger.c
```diff
@@ -217,12 +216,27 @@ void logger_stop ()
...
 	static time_t now;                   //for holding system time
 	static char *time_str;               //for string representation of system time
 	static char final_msg[MAX_LOG_LEN];  //final message to be logged to requested locations
 	static int len;                      //holding lengths of various strings
+	static char msg[MAX_LOG_LEN - 100];  //holds the expanded format string (100 to make room for log header)
+	va_list args;                        //this holds the variable-length argument list
+	
+	//expands the format string into msg
+	va_start(args, format);
+	vsprintf(msg, format, args);
```
AVSurfer123, line 238: probably should put this after you check if the level is high enough

shm_wrapper/shm_wrapper.c
```diff
@@ -278,26 +265,31 @@ void shm_init (process_t process)
 		
 		//mutual exclusion semaphore, catalog_mutex with initial value = 0
 		if ((catalog_sem = sem_open(CATALOG_MUTEX_NAME, O_CREAT, 0660, 0)) == SEM_FAILED) {
-			error("sem_open: catalog_mutex@dev_handler");
+			log_printf(FATAL, "sem_open: catalog_mutex@dev_handler");
```
AVSurfer123, line 268: For some of these, I see you don't include the `strerror(errno)` so just make sure you add those to all the system C functions.

logger/logger.c
benliao1: Good point. Fixed

shm_wrapper/shm_wrapper.c
benliao1: 👌

Issue #54, opened 2020-07-17
[EXECUTOR] Stopping mode while sleeping causes error


AVSurfer123:
Stopping the mode subprocess while Python is sleeping causes a segfault and more importantly doesn't notify the student that the function is taking too long. Need to fix this, or find an alternative signal to sending an asynchronous exception.

AVSurfer123:
Also,
- calling Robot.run in main function causes executor to become unresponsive.
- calling set_value with NameError causes buffer overflow

Issue #50, opened 2020-07-13
[STUDENTAPI] Add way for student code to access the starting position of the robot


AVSurfer123:
Need to add an API function that will simply pull the starting position from aux SHM

Issue #44, opened 2020-07-11
[EXECUTOR] Add exception handling in the student API for missing devices and gamepad


AVSurfer123:
Have a check/exception when device is not found in SHM, depends on #43

Issue #39, opened 2020-07-08
[EXECUTOR] Parse student code to determine subscription bitmap


benliao1:
Executor can parse student code and determine which parameters of which devices are used in student code, and set the subscription bitmap for the executor appropriately. This ensures that the device handler will fetch these values from the indicated devices and supply them to the executor / student code.

If the student code indicates for a device that does not exist and the code is running in `TELEOP` or `AUTO` mode, executor should issue a warning to the student code that the indicated device is not attached, but not crash the code. Rather, it should keep running, and return some sort of default value for those values.

AVSurfer123:
This seems way harder to implement than the benefit gained by it. It's even harder to implement because we are using C and the string parsing sucks.

Also I just checked the SHM code and it seems that the executor currently has no way to determine if the UID doesn't exist in SHM since the SHM don't return an error status. So executor would just be giving garbage value on read.

However, the executor should definitely crash if it can't find the device. It would be very bad to catch that exception as it's the only way for the student to know that their devices aren't connected/working. We should only catch errors that are caused by us.

benliao1:
It's actually a really key part if we want to reduce the amount of computation that the device handler does by a significant factor. Take, for example, the KoalaBear, which may end up having 25+ parameters. Without the subscription bitmap, the device handler runtime would (be default) subscribe to all of these parameters. This means that the dev handler would need to decode all 25 of these parameters 10, 20 or 30 times a second (depending on how quickly the Arduino can run, and our requested subscription). But the students are almost certainly not going to be using all 25 of these parameters (perhaps they only use 2--the duty cycles on both connected motors), in which case the device handler can support the student code while doing <5% of the original work it was doing. 

I do agree that we need to figure out exactly what the load on the device handler is before we implement it though, and that this part of it in particular is going to be very difficult--which is why it'll be in v1.1 if it's implemented at all.

Regarding what happens when a student specifies an unregistered device, right now it simply crashes the student code without returning the robot mode to `IDLE` right?

AVSurfer123:
No like I said the student API has no way of knowing whether the device actually exists since SHM doesn't give any return status. It just silently returns after it catches the exception. So device_write does nothing if the device doesn't exist, and device read will return garbage values from what it malloced. I should add an issue to fix that, but its pretty simple. You just need to change the return type from void to int.

AVSurfer123:
[...]

PR #61, opened 2020-07-20
[EXECUTOR] Added lots of enhancements to the executor that interface with SHM and device handler

### LINKED ISSUE #33: [EXECUTOR] Stop all motors when changing to IDLE mode
The executor needs to tell the dev_handler to set all motors to 0 power whenever the mode becomes IDLE to ensure safety.

### LINKED ISSUE #46: [ORGANIZATION] Make a build system and CI system
We should have 1 global makefile that calls the smaller makefiles. Also start integrating a CI/CD system like Travis that will automatically run tests and make sure that Runtime builds and runs.

AVSurfer123:
Closes #33, #39, #44, #50, Fixed buffer overflow bug in #54 .

Adds several enhancements to `executor`:
- Adds exception handling for missing/invalid devices
- Adds subscription requests for devices by parsing code
- Adds exception handling for actions
- Adds reset to SHM on robot IDLE

Also:
- Made `logger_stop` and `shm_stop` be called with `atexit` so that `exit` will automatically close those libraries. We should no longer call those stop functions explicitly
- Added `print_sub_map` to print the current subscription map

And this PR also includes PR #56 which closes #46 since Docker would best work with the updated executor.


benliao1:
Should we link issue #54 to this as well?

AVSurfer123:
No this doesn't fix that bug

benliao1:
Ok~~ then we should probably remove it from the first commit on the pull request

AVSurfer123:
For #54, if the main loop is in `time.sleep` and you cancel it, the mode process will segfault but that won't crash the parent process so you might not notice it. And to break the code parser you can look at my comment on #39.

PR #65, opened 2020-07-22
[SHM][NET_HANDLER] Added subscription requests to net_handler

### LINKED ISSUE #40: [NET_HANDLER] Implement interfacing with shm subscription bitmap
Net handler should write to the subscription bitmap once Dawn supports collapsible parameters in their dashboard.

This ensures that the UDP connection can read out the requested device parameters from the student, and that the data will be the latest values of the parameters as updated by the device handler.

AVSurfer123:
Closes #40 
- Changed get_sub_requests to be different depending on the calling process
- Made net_handler subscriptions be default on, so that student can at first see all devices
- Added DevData processing code to tcp_conn which places sub requests
- Added check for the current sub requests when sending data in udp_conn
- Added test code in tcp_client to turn polarbear data off and on

### DIFF
--- a/net_handler/tcp_conn.c
+++ b/net_handler/tcp_conn.c
@@ -254,7 +254,25 @@ static int recv_new_msg (int conn_fd, int challenge_fd)
 			log_printf(ERROR, "Cannot unpack dev_data msg");
 			return -2;
 		}
-		log_printf(WARN, "Received device data msg which has no implementation");
+		for (int i = 0; i < dev_data_msg->n_devices; i++) {
+			Device* req_device = dev_data_msg->devices[i];
+			device_t* act_device = get_device(req_device->type);
+			if (act_device == NULL) {
+				log_printf(ERROR, "Invalid device subscription: device type %d is invalid", req_device->type);
+				continue;
+			}
+			uint32_t requests = 0;
+			for (int j = 0; j < req_device->n_params; j++) {
+				if (req_device->params[i]->val_case == PARAM__VAL_IVAL) {
+					// log_printf(DEBUG, "Received device subscription for %s param %s", act_device->name, act_device->params[j].name);
+					requests |= (req_device->params[j]->ival << j);
+				}
+			}
+			int err = place_sub_request(req_device->uid, NET_HANDLER, requests);
+			if (err == -1) {
+				log_printf(ERROR, "Invalid device subscription: device uid %llu is invalid", req_device->uid);
+			}
+		}
 		//TODO: tell UDP suite to only send parts of data, not implemented by Dawn yet
 		
 		dev_data__free_unpacked(dev_data_msg, NULL);
--- a/net_handler/tcp_conn.c
+++ b/net_handler/tcp_conn.c
@@ -1,4 +1,5 @@
 #include "tcp_conn.h"
+#include "pbc_gen/device.pb-c.h"
 
 //used for creating and cleaning up TCP connection
 typedef struct {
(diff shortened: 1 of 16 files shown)


net_handler/tcp_conn.c
```diff
@@ -254,7 +254,25 @@ static int recv_new_msg (int conn_fd, int challenge_fd)
...
+		for (int i = 0; i < dev_data_msg->n_devices; i++) {
+			Device* req_device = dev_data_msg->devices[i];
+			device_t* act_device = get_device(req_device->type);
+			if (act_device == NULL) {
+				log_printf(ERROR, "Invalid device subscription: device type %d is invalid", req_device->type);
+				continue;
+			}
+			uint32_t requests = 0;
+			for (int j = 0; j < req_device->n_params; j++) {
+				if (req_device->params[i]->val_case == PARAM__VAL_IVAL) {
```
benliao1, line 266: I think I told Dawn to use the boolean to request subscriptions (so if student requests param `i` then the value of param `i` coming in will be `True`, and vice versa for `False`). You can assume that any nonexistent params or params that aren't specified in the message are `False`.

shm_wrapper/shm_wrapper.c
```diff
@@ -535,10 +536,16 @@ void device_disconnect (int dev_ix)
...
+	//reset bitmap values to 0
 	dev_shm_ptr->cmd_map[0] &= (~(1 << dev_ix));   //reset the changed bit flag in cmd_map[0]
 	dev_shm_ptr->cmd_map[dev_ix + 1] = 0;          //turn off all changed bits for the device
 
+	dev_shm_ptr->exec_sub_map[0] |= 1 << dev_ix;
+	dev_shm_ptr->exec_sub_map[dev_ix + 1] = -1;
+
+	dev_shm_ptr->net_sub_map[0] |= 1 << dev_ix;
+	dev_shm_ptr->net_sub_map[dev_ix + 1] = -1;
+
```
benliao1, line 548: So correct me if I'm wrong but on device disconnect we reset the subscriptions to all parameters for that device index?

levincent06: If a device with shared memory index `i` disconnects and we reset the subscriptions to `-1` , then the next device assigned to the same shm index will start with a subscription to all params right?

I guess that would be safer in the sense that we have all possible incoming data immediately and we tell dev handler what data we don't need. But I also think setting them to all 0 would save data encoding time... though not by much because net handler and executor will probably send updated subscription requests within a second, right?

AVSurfer123: So net_handler has to always reset to -1 since otherwise Dawn would never know what params exist/can be selected. Unless we are assuming that they also have a copy of the devices within `runtime_util.c` . I'm note sure if they should since their API should be as abstract as possible to allow any device and any param.

net_handler/tcp_conn.c
AVSurfer123: Ok changed it to check the boolean value instead.

shm_wrapper/shm_wrapper.c
benliao1: Ok that makes sense. We should make a note that once Dawn implements the subscriptions, we should change this back to setting to 0 by default, since then Dawn will have the ability to change this bitmap.

AVSurfer123: No regardless it needs to be -1. How would dawn know what parameters the device has and to subscribe to?

benliao1: @AVSurfer123  Since all the values in the protobufs are optional, by default we can be sending DevData messages to Dawn that don't have any of the `ival` `fval` or `bval` fields set. We only send Dawn the list of connected devices (with name, uid, and type), and then each connected device will have a list of all of their parameters (which only contain names). That's enough information for Dawn to display the dashboard of all devices with all parameter names collapsed.

AVSurfer123: oh right now if you don't subscribe you don't get the param name, because the name is part of the Param protobuf struct

PR #91, opened 2020-08-05
[LOWCAR] Verifies lowcar devices have same or greater functionality as before

### LINKED ISSUE #31: [LOWCAR] update and test lowcar code to function properly with Device Handler
Currently it is not known if the Device Handler is implementing the protocol exactly the same way as in the old runtime (it doesn't need to be, we are free to design the protocol however we want). However, any changes made to the protocol during the design of the current Device Handler must be reflected in the lowcar (Arduino device) code.

Additionally, with the addition of the LOG packet, other parts of the lowcar code can finally be debugged and verified. This should be done, and any changes or bugs found should be noted and fixed.

benliao1:
Verifies that the following devices work as well as with the old Runtime, or better:
- BatteryBuzzer
- KoalaBear
- LimitSwitch
- LineFollower
- PolarBear
- ServoControl

Also verifies that `dev_handler` can handle 10+ devices no problem (10 devices resulted in about 22% CPU usage across all Runtime processes!)

Adds a `lowcar_test.py` but this will be deleted in `docs_issue_68` branch; it just might be useful in the future so it should be in the history of `master` somewhere.

Deleted `lib/Encoder` folder, because we don't need it any more.

Did do some style fixes, but did not finish. Will finish on `docs_issue_68` branch.

Closes #31 (except for KoalaBear PID and encoder features)


lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -5,358 +5,346 @@
...
+    
+    // initialize encoders
+    enc_a = enc_b = 0.0;
+    attachInterrupt(digitalPinToInterrupt(AENC1), handle_enc_a_tick, RISING); // set interrupt service routines for encoder pins
+    attachInterrupt(digitalPinToInterrupt(BENC1), handle_enc_b_tick, RISING);
+    
+    // initialize PID controllers
+    this->pid_a = new PID();
+    this->pid_b = new PID();
+    this->pid_enabled_a = this->pid_enabled_b = FALSE; // TODO: change to true when PID written
```
AVSurfer123, line 95: why is FALSE not `false`? Also, they should default to disabled and the students should turn them on when they wish.

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -5,358 +5,346 @@
...
-			return sizeof(float);
+KoalaBear::KoalaBear() : Device(DeviceType::KOALA_BEAR, 13) {
+    // initialize motor values    
+    this->target_speed_a = this->target_speed_b = 0.0;
+    this->deadband_a = this->deadband_b = 0.05;
+    
+    // initialize encoders
+    enc_a = enc_b = 0.0;
+    attachInterrupt(digitalPinToInterrupt(AENC1), handle_enc_a_tick, RISING); // set interrupt service routines for encoder pins
+    attachInterrupt(digitalPinToInterrupt(BENC1), handle_enc_b_tick, RISING);
```
AVSurfer123, line 90: I think we'll eventually need 4 total interrupts

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -5,358 +5,346 @@
...
+    } else {
+        target_B = this->target_speed_b;
+    }
+
+    digitalWrite(SLEEP, HIGH);
+    digitalWrite(RESET, LOW);
+
+    // send target duty cycle to drive function
+    drive(target_A * -1.0, MTRA); // by default, MTRA drives the wrong way
+    drive(target_B, MTRB);
```
AVSurfer123, line 291: The motor directions should be device parameters

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -5,358 +5,346 @@
...
+    
+    delay(1); // delay 1 millis to slow down the loop
```
AVSurfer123, line 293: remove this probably

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -365,24 +353,24 @@ int KoalaBear::sign (float x) {
...
+    int currpwm1 = 255;
+    int currpwm2 = 255;
+
+    // Determine how much to move the pins down from 255
+    // Sanity check: If |target| == 1, move max speed --> pwm_difference == 255
+    //				 If |target| == 0, stop moving	  --> pwm_difference == 0
+    int pwm_difference = (int)(target * direction * 255.0); // A number between 0 and 255 inclusive (255 is stop; 0 is max speed);
+
+    if (direction > 0) { // Moving forwards
+        currpwm1 -= pwm_difference;
```
AVSurfer123, line 369: why isn't this +=

lowcar/devices/KoalaBear/PID.cpp
```diff
@@ -1,125 +1,61 @@
...
+    this->integral += error * (interval_secs); //compute the new value of this->integral using right-rectangle approximation
+    
+    // store the current values into previous values for use on next loop
+    this->prev_error = error;
+    this->prev_pos = curr_pos;
+    this->prev_desired_pos = desired_pos;
+    this->prev_time = curr_time;
+    
+    // return kp * error + ki * integral of error * kd * "derivative" of error
+    return (this->kp * error) + (this->ki * this->integral) + (this->kd * (error - this->prev_error));
```
AVSurfer123, line 31: you need to divide the derivative term by `interval_secs`. you also are using the new `prev_error`, which is just equal to `error` so that won't work.

lowcar/devices/KoalaBear/KoalaBear.h
```diff
@@ -1,35 +1,40 @@
 #ifndef KOALABEAR_H
 #define KOALABEAR_H
 
+#include <SPI.h>
```
AVSurfer123, line 4: why is this needed? I don't see any SPI devices

lowcar/devices/LineFollower/LineFollower.cpp
```diff
@@ -1,24 +1,28 @@
...
 
-size_t LineFollower::device_read (uint8_t param, uint8_t *data_buf) {
-	//use data_ptr_float to shove 10-bit sensor reading into the data_buf (even though data_buf is of type uint8_t)
-	float *data_ptr_float = (float *)data_buf;
-	*data_ptr_float = ((float) analogRead(this->pins[param])) / 1023;
-	return sizeof(float);
+size_t LineFollower::device_read(uint8_t param, uint8_t *data_buf) {
+    // use data_ptr_float to shove 10-bit sensor reading into the data_buf (even though data_buf is of type uint8_t)
+    float *data_ptr_float = (float *)data_buf;
+    *data_ptr_float = 1.0 - ((float) analogRead(this->pins[param])) / 1023;
```
AVSurfer123, line 19: better to do 1023.0

lowcar/devices/PolarBear/PolarBear.cpp
```diff
@@ -1,96 +1,77 @@
...
+    MOTOR_CURRENT = 1,
+    DEADBAND = 2
 } param;
 
-PolarBear::PolarBear () : Device (DeviceType::POLAR_BEAR, 2) {
-	this->duty_cycle = 0;
-	this->deadband = 0.05;
-
-	this->dpwm_dt = 255 / 200000;
+PolarBear::PolarBea () : Device(DeviceType::POLAR_BEAR, 2) {
```
AVSurfer123, line 10: typo: PolarBear

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: To answer the first question, it's because `pid_enabled_a` is a boolean, but in C++, true/false is not necessarily 1 and 0. So if you use the C++ true/false values, when you send them over serial to `dev_handler`, the true value ends up being a random positive number between 1 and 255 since C doesn't have a builtin boolean type. The solution is to define two constants, TRUE as 1 and FALSE as 0, so that when we send them to `dev_handler` they can be treated as true/false values in C.

To answer the second point, I talked a bit about this with Jesse and Jonathan. They both believed that our students likely won't be advanced enough to be able to tune the PID controllers themselves. So, by default these will be enabled and the KP, KI, and KD values will have default values that we find for them. They can change the tuning constants and/or disable the PID control if they wish, but the PID control is mostly for us to try and create a better experience for the students (robots actually drive straight in autonomous, etc.)

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: I talked to Sherwin about this, because I was confused as well (he told me how to program this). We won't need 4 total interrupts, because we each encoder is connected to two pins; one is the pin to interrupt on and the other is the pin to read which direction the motor is going. I can explain on Saturday, it's hard to explain in writing.

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: Yes, forgot about this. This was added when I was trying to solve a bug with the LEDs....

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: It has to do with the controls for the motors... basically if both control pins are high, the motor is stopped; lowering the voltage on one pin while keeping the other one high moves the motor one way (and the lower the voltage is, the faster it goes); lowering the voltage on the other pin while keeping the first pin high moves the motor the other way.

lowcar/devices/KoalaBear/PID.cpp
benliao1: Ah, you're right. Lemme fix this

lowcar/devices/PolarBear/PolarBear.cpp
benliao1: o god how did I not see this one / how did it compile??

lowcar/devices/KoalaBear/KoalaBear.h
benliao1: You're right, it's not needed any more. I think it used to be used for the current limiting on the PolarBear or something, I'm not sure though.

executor/lowcar_test.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # if switch0 != prev_switch0:
+    #     print(f"switch0 switched to {switch0}")
+    #     prev_switch0 = switch0
+    # if switch1 != prev_switch1:
+    #     print(f"switch1 switched to {switch1}")
+    #     prev_switch1 = switch1
+    # if switch2 != prev_switch2:
+    #     print(f"switch2 switched to {switch2}")
+    #     prev_switch2 = switch2
```
levincent06, line 53: GitHub is warning me that there's no newline at the end of this file :P

executor/lowcar_test.py
```diff
@@ -0,0 +1,53 @@
+PolarBear = '12_2541551405711093506' # flashed with -t
+KoalaBear = '13_1250999896333' # 123456789
+Switch = '0_2882400000' # ABCDEF
+ServoControl = '7_10477124133121' # 987654321
```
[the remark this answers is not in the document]
benliao1: 👌

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -5,358 +5,346 @@
...
+    PID_KP_B      = 12,
+    PID_KI_B      = 13,
+    PID_KD_B      = 14,
+    ENC_B         = 15
 } param;
 
 typedef enum {
-	MTRA = 1,
-	MTRB = 0,
+    MTRA = 1,
```
levincent06, line 56: Do you happen to know why this ordering is irregular?

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -5,358 +5,346 @@
...
+            float_buf[0] = this->pid_a->get_kp();
+            return sizeof(float);
+        case PID_KI_A:
+            float_buf[0] = this->pid_a->get_ki();
+            return sizeof(float);
+        case PID_KD_A:
+            float_buf[0] = this->pid_a->get_kd();
+            return sizeof(float);
+        case ENC_A:
+            int_buf[0] = enc_a;
```
levincent06, line 130: I think we should have `this->enc_a` and `this->enc_b` this in file for clarity and consistency

lowcar/devices/PolarBear/PolarBear.cpp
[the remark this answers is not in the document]
levincent06: Ashwin beat me to it lol

lowcar/devices/LineFollower/LineFollower.cpp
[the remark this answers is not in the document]
benliao1: Yea, I think it was reading white high and black low previously; now it's doing the opposite

lowcar/devices/PolarBear/LED.cpp
```diff
@@ -1,75 +1,57 @@
...
-	digitalWrite(LED_RED,HIGH);
-	digitalWrite(LED_YELLOW,HIGH);
-
-	delay(1000);
-
-	digitalWrite(LED_GREEN,LOW);
-	digitalWrite(LED_RED,LOW);
-	digitalWrite(LED_YELLOW,LOW);
+// place all LED control functions in here
+void ctrl_LEDs(float duty_cycle) {
```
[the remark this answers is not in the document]
benliao1: Yeah that's what it's doing. I'm going to change it to match the KoalaBear ones though, like you mentioned in the comment below.

lowcar/devices/KoalaBear/LEDKoala.cpp
```diff
@@ -18,16 +18,16 @@ LEDKoala::LEDKoala () {
 }
 
 //place all LED control functions in here
-void LEDKoala::ctrl_LEDs (float vel, bool enabled) {
-	ctrl_red(vel, enabled);
-	ctrl_yellow(vel, enabled);
-	ctrl_green(vel, enabled);
+void LEDKoala::ctrl_LEDs (float vel, float deadband, bool enabled) {
```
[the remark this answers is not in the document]
benliao1: Yeah style fixes can be left to `docs_issue_68` but I will change the PolarBear ones to match these ones

lowcar/devices/PolarBear/PolarBear.cpp
```diff
@@ -101,22 +82,22 @@ int PolarBear::sign(float x) {
...
-	delayMicroseconds(1 / this->dpwm_dt); // About 784
+    float direction = (target > 0.0) ? 1.0 : -1.0;
+    int currpwm1 = 255;
+    int currpwm2 = 255;
+
+    // Determine how much to move the pins down from 255
+    // Sanity check: If |target| == 1, move max speed --> pwm_difference == 255
+    //               If |target| == 0, stop moving    --> pwm_difference == 0
+    int pwm_difference = (int)(target * direction * 255.0); // A number between 0 and 255 inclusive (255 is stop; 0 is max speed);
+
```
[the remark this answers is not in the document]
benliao1: Yes, ok will add comments

lowcar/devices/KoalaBear/PID.h
```diff
@@ -2,37 +2,31 @@
...
-	float readPos();
-	float readVel();
-	void updateVel();
-	void updatePos();
+	PID();
+	float compute(float curr_pos);
+        
+	void set_coefficients(float kp, float ki, float kd);
+	void set_target_speed(float new_target_speed);
+    void set_pos_to(float curr_pos);
```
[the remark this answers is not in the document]
benliao1: 👌

lowcar/devices/KoalaBear/PID.cpp
```diff
@@ -1,125 +1,61 @@
...
+float PID::get_kp() { return this->kp; }
+float PID::get_ki() { return this->ki; }
+float PID::get_kd() { return this->kd; }
+
+// *********************** HELPER FUNCTIONS *********************** //
+
+// converts speed in units of duty cycle (-1.0 to 1.0)
+// to units of encoder ticks per second, tps
+float PID::duty_cycle_to_tps(float duty_cycle) {
+    return 300.0 * duty_cycle; // TODO: determine this function. will probably be of the form y = kx (if not can go do a regression on a calculator after some tests)
```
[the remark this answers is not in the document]
benliao1: Idk, I've never gotten to work with a motor encoder in a testable way so I'm not sure. My guess is that they should be pretty consistent if it's the same type of motor?

But I guess we'll find out when the motors arrive at my house lol

executor/lowcar_test.py
benliao1: I'm gonna delete this file from the PR, per Ashwin's suggestion, so we don't need to worry about this

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: No, idk.... I'll reverse it

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: So `enc_a` and `enc_b` are actually not class fields, so there is no variable `this->enc_a` or `this->enc_b`. The way those interrupt handlers work is that they have to be declared statically, outside of any class, which is why the two interrupt handlers aren't like `KoalaBear::handle_enc_a_tick`. Sine those functions are declare outside of any class, they can't use any class variables. So `enc_a` and `enc_b` are essentially global variables on a KoalaBear.

lowcar/devices/KoalaBear/PID.cpp
```diff
@@ -1,125 +1,61 @@
...
-float PID::readVel() {
-  return vel;
-}
-
-void PID::updatePos() {
-  pos = enc->read();
+ * Arguments:
+ *    curr_pos: current value of the encoder, as a float
+ */
+float PID::compute(float curr_pos) {
```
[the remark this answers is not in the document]
benliao1: 👌

lowcar/devices/KoalaBear/KoalaBear.h
benliao1: Oh no never mind, this is actually used in KoalaBear. If you look at the setup functions from electrical, you'll see SPI functions being used....

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: 1.0 feature? Since this would require an additional parameter in `runtime_util.c`?

Let's stick with this for now, I'll add this as an issue though that I can add when working with the encoders (I'm pretty sure that the PID math doesn't take this into account as of now either)

AVSurfer123: I mean, KoalaBear needs a revamp anyways, so this can be part of that issue #92

lowcar/devices/KoalaBear/KoalaBear.cpp
AVSurfer123: Ok I understand

lowcar/devices/KoalaBear/KoalaBear.cpp
AVSurfer123: even if that were the case, you probably wouldn't want the pid to run during teleop. also right now theres no way for them to  give a setpoint to the pid controller in the device params, which is pretty important.

lowcar/devices/KoalaBear/PID.cpp
```diff
@@ -17,18 +17,20 @@ PID::PID() {
 float PID::compute(float curr_pos) {
     unsigned long curr_time = micros(); // get the current time
     float interval_secs = ((float)(curr_time - this->prev_time)) / 1000000.0; // compute time passed between loops, in seconds
-    float desired_pos = prev_desired_pos + (duty_cycle_to_tps(this->target_speed) * interval_secs); // compute the desired position at this time
+    float desired_pos = this->prev_desired_pos + (duty_cycle_to_tps(this->target_speed) * interval_secs); // compute the desired position at this time
```
AVSurfer123, line 20: Not sure if it should say `curr_pos` instead of `this->prev_desired_pos` but I guess you can test that when you get the encoders.

PR #97, opened 2020-08-12
[EXECUTOR] Split challenges into new file, added challenge name file

### LINKED ISSUE #84: [EXECUTOR] Have the list of challenges be in a separate file, so that tests can modify it
It would be nice to have the list of challenge functions be a separate file that is read by `executor` when the mode begins. This would allow the test framework to test the CHALLENGE mode more easily.

AVSurfer123:
- Made the student challenge code file another CLI input to the executor process. By default it is now `challenges` but we can make this `studentcode` for now if Dawn isn't going to split the two files on their end. 
- Required a `CHALLENGES` list in the Python file that contains the names of the challenges to run

To do this, needed to move the Protobuf unpacking logic for the challenge inputs into the executor, since the number of challenges is now variable. 

Closes #84


tests/client/executor_client.c
```diff
@@ -23,33 +20,39 @@ void start_executor (char *student_code)
...
+			new_python_path = malloc(strlen(PYTHONPATH) + 1 + strlen(path_to_test_student_code) + 1);
 			new_python_path[0] = '\0';
-			strcat(new_python_path, python_path_header);
+			strcat(new_python_path, PYTHONPATH);
+			strcat(new_python_path, "=");
 			strcat(new_python_path, path_to_test_student_code);
-		} else if (strstr(python_path, "tests/student_code") == NULL) { //if PYTHONPATH is defined but the path we want isn't there
-			new_python_path = malloc(strlen(python_path) + strlen(path_to_test_student_code) + strlen(python_path_header) + 1);
+		} else { // if PYTHONPATH is defined
+			new_python_path = malloc(strlen(PYTHONPATH) + 1 + strlen(python_path) + 1 + strlen(path_to_test_student_code) + 1);
```
benliao1, line 31: you actually don't need the intermediate +1s because `strcat` will concatenate the two strings together, with the first character of the second string starting at the location of the terminating null character of the first string. But this should still be fine because we're allocating more memory than needed

PR #101, opened 2020-08-14
Converted test framework to use Cython tests

### LINKED ISSUE #90: [TEST] Rewrite test.c in Cython and integration tests in Python
The test framework checking would be much nicer if the tests and outputs were written in Python and we check these outputs with the results from calling the C test clients.

AVSurfer123:
Converted the test output checker to a Cython module `tester.pyx` and converted all the test cases to Python. This allows for much easier checking which doesn't care about whitespace. It also allows easier enhancements in the future.

Also
- modified `test.sh` for the new framwork
- added a root Makefile with build, run, clean, and test targets

Closes #90 

WIP:
Waiting on `challenge_files` branch to get merged to use the new `send_challenge_data` and `start_executor` functions
Waiting on `dev_handler_cli` branch to add new client to the test framework


benliao1:
Test framework was revamped during the Fa '20 semester by @levincent06 and @DanielMolina24 to make it easier to use!

PR #103, opened 2020-08-19
[TEST] Changed test.c to use thread, not process. Remove temp file

benliao1:
As per title of pull request. Removes the need of a temp file in test framework.

### DIFF
--- a/tests/test.c
+++ b/tests/test.c
@@ -1,11 +1,8 @@
 #include "test.h"
 
-#define TEMP_FILE "/tmp/temp_logs.txt"
-
 #define DELIMITER_WIDTH 80
 
-pid_t output_redirect;          // holds process ID of output redirection process
-FILE *temp_fp;                  // file pointer of open temp file
+pthread_t output_handler_tid;   // holds thread ID of output handler thread
 int save_std_out;               // saved standard output file descriptor to return to normal printing after test
 int pipe_fd[2];                 // read and write ends of the pipe created between the parent and child processes
 char *test_output = NULL;       // holds the contents of temp file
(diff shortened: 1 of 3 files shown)


tests/test.c
```diff
@@ -1,61 +1,53 @@
...
+ * It constantly reads from the read end of the pipe, and then prints it to stderr
+ * and stores in test_output. It is canceled by the test in the function end_test().
+ * Arguments:
+ *    args: always NULL
+ * Returns: NULL
+ */
+static void *output_handler(void *args) {
+    size_t curr_size = 1024;
+    size_t num_total_bytes_read = 0;
+    char *curr_ptr = test_output = malloc(curr_size);
```
AVSurfer123, line 24: bad style. we should never do multiple assignments on same line

tests/test.c
```diff
@@ -1,61 +1,53 @@
...
+/**
+ * This thread prints output to terminal and also copies it to standard output
+ * It constantly reads from the read end of the pipe, and then prints it to stderr
+ * and stores in test_output. It is canceled by the test in the function end_test().
+ * Arguments:
+ *    args: always NULL
+ * Returns: NULL
+ */
+static void *output_handler(void *args) {
+    size_t curr_size = 1024;
```
AVSurfer123, line 22: should be MAX_LOG_LEN

tests/test.c
```diff
@@ -83,98 +75,66 @@ static void fprintf_delimiter(FILE *stream, char *format, ...) {
...
-
-        // kill the child on Ctrl-C (SIGINT)
-        signal(SIGINT, kill_child);
-
-        // register the clean up handler
-        atexit(cleanup_handler);
+    // save the standard out attached to the terminal for later
+    save_std_out = dup(fileno(stdout));
+    
+    // route stdin to read end of pipe
```
AVSurfer123, line 104: comment should be other way around. read end gets routed to stdin

tests/test.c
```diff
@@ -83,98 +75,66 @@ static void fprintf_delimiter(FILE *stream, char *format, ...) {
...
+    if (dup2(pipe_fd[0], fileno(stdin)) == -1) {
+        fprintf(stderr, "dup2 stdin to read end of pipe: %s\n", strerror(errno));
+    }
+    // route stdout to write end of pipe
```
AVSurfer123, line 108: comment should be other way around

tests/test.c
benliao1: 👌

tests/test.c
benliao1: 👌

tests/test.c
benliao1: 👌

tests/test.c
benliao1: 👌

PR #105, opened 2020-08-19
[MISC] Put scripts into scripts/ folder, added runtime.sh

benliao1:
As per title of pull request.


scripts/test.sh
```diff
@@ -68,17 +68,15 @@ function run_tests {
 trap 'sigint_handler' INT
 
 # build all of Runtime
-./build.sh
+./scripts/build.sh
```
benliao1, line 71: This will fail now that this file doesn't exist any more.... does replacing it with `runtime build` work?

runtime
```diff
@@ -0,0 +1,52 @@
...
+
+    format)
+        ;;
+
+    clean)
+        pushd shm_wrapper && make clean && popd
+        pushd net_handler && make clean && popd
+        pushd executor && make clean && popd
+        pushd dev_handler && make clean && popd
+        pushd tests && make clean && popd
```
benliao1, line 45: Do you think we should add like `rm /tmp/log-fifo && rm /tmp/challenge.sock && rm /tmp/ttyACM* && rm /dev/shm` to this?

runtime
```diff
@@ -0,0 +1,52 @@
...
+
+# This file is meant to be sourced by the user, who can then run commands corresponding
+# to the name of the script to be run, e.g.
+#     runtime build                      (to run build.sh)
+#     runtime run                        (to run run.sh)
+#     runtime test integration/tc_68_2   (to run the automated test tc_68_2)
+
+# To source this file to make this function available, issue the command
+#     source runtime.sh
+
```
benliao1, line 11: Update these comments... how does it work now? Do you not have to put `./` in front of it?

runtime
```diff
@@ -0,0 +1,52 @@
...
+        ;;
+
+    clean)
+        pushd shm_wrapper && make clean && popd
+        pushd net_handler && make clean && popd
+        pushd executor && make clean && popd
+        pushd dev_handler && make clean && popd
+        pushd tests && make clean && popd
+        ;;
+
```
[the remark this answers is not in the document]
AVSurfer123: i think it would be better to have it here. I'll just have to modify some of the paths in the flash script but shouldn't be an issue

scripts/test.sh
AVSurfer123: ah yes. i'll need to replace it with `./runtime build`

runtime
AVSurfer123: the last one is dangerous if other SHM is being used by the OS. I think we can do the first ones though

runtime
AVSurfer123: you don't have to add the ./ if you add it to your PATH. and I also made it callable from anywhere

PR #107, opened 2020-08-19
[MISC] Fixed shared memory, logger, and test framework to conform to style guide

benliao1:
Also cleaned up interface for shared memory and logger (removed `shm_stop()`, `logger_stop()`, etc.).


benliao1:
So the build on this branch deletes `tcp_client` and `udp_client`, which means that Travis no longer does the sanity check running `net_handler` with `tcp_client` and `udp_client`. But it should be fine, since that check and more are covered into the new integration test cases 16, 17, and 18.

PR #122, opened 2020-10-10
[MISC] Gamepad error check

AVSurfer123:
While working on Dawn, I found that the current Gamepad connection check is flawed and so I fixed it.

Also, having a nonzero uid for the CustomData is easier since the uid 0 is the default Protobuf value corresponding to 0 bytes and makes life really annoying in Dawn.


benliao1:
Approved and passing Travis tests...

benliao1 (approved):
Looks good! Changes make sense~~

Issue #78, opened 2020-07-28
[TEST] Allow for more sophisticated test output matching


benliao1:
Right now, the test framework only supports exact character-by-character string matching. This can be annoying (because whitespace characters have to match exactly for the test to pass) or make the test less effective (you can't compare across logs, for example, because the timestamps in the logs are guaranteed to be different each time you run the test).

It would be very useful to be able to do the following in output match:
- put some kind of "whitespace wildcard" like `[:space:]` from regex or `%ws` escape sequence to indicate "some amount of whitespace should appear here in the output"
- put some kind of "number wildcard" like `[:digit:]` from regex or `%d` escape sequence to indicate "some sequence of digits should appear here in the output"

Could be implemented in a few different ways:
- Doing some basic string parsing in C
- Calling a short Python script to do the parsing for us

benliao1:
My concerns about complexity: keeping in line with Runtime's goal of not using many third-party libraries, the goal is for this issue to not involve including a third-party library to handle the regular expressions for us. Additionally, our tests do not require us to use the entire functionality of regex to search through our test output; it's simply not necessary for our use cases (regex is typically used for find-and-replace algorithms and tools; we are simply comparing output of a program with what is expected).

This is not to say that we cannot borrow regex syntax for our use, for example using `[:digit:]` and `[:space:]` would be useful for Runtime members to learn regex in the future.

Additionally, if we restrict the set of "wildcard" sequences to just `[:digit:]` and `[:space:]` and implement it ourselves, it is more fun for the person writing it and easier for someone onboarding to look at the code and understand what's going on (without having to comb through pages and pages of potentially inadequate documentation).

PR #123, opened 2020-10-18
Regex tests

0neDividedbyZer0:
Created Regex test framework. Uses the regex.h library in C, and has flags to allow tests to try and use regex or just do string detection. Updated tc_71 to have NO_REGEX flags for compatibility.


tests/test.h
```diff
@@ -1,11 +1,25 @@
 #ifndef TEST_H
 #define TEST_H
 
+#include <regex.h> 
+#include <string.h>
+#include <stdio.h>
```
DanielMolina24, line 6: You don't need to include stdio.h since thats already included in the clients below

tests/test.h
```diff
@@ -1,11 +1,25 @@
 #ifndef TEST_H
 #define TEST_H
 
+#include <regex.h> 
+#include <string.h>
```
DanielMolina24, line 5: You dont need to include string.h since thats already included in the clients below

tests/test.h
DanielMolina24, line 6: Same as comment above

tests/test.h
```diff
@@ -1,11 +1,25 @@
...
+#include <stdio.h>
+#include <stddef.h>
```
DanielMolina24, line 7: Can you add a comment on what the studded is being used for?

tests/test.h
```diff
@@ -1,11 +1,25 @@
...
+
 #include "client/dev_handler_client.h"
 #include "client/executor_client.h"
 #include "client/net_handler_client.h"
 #include "client/shm_client.h"
 
+static const uint8_t NO_REGEX = 0;
+static const uint8_t REGEX = 1;
```
DanielMolina24, line 15: This one is just a nit but I think it might be cleaner to have these in an enum

tests/test.c
```diff
@@ -138,10 +138,32 @@ void end_test() {
     printf("************************************** Running Checks... ******************************************\n");
 }
 
+//Implementation of strstr but with a regex string for needle
+char *rstrstr(const char *haystack, const char *needle) {
+    regex_t expr;
+    regmatch_t tracker;
+
+    if (regcomp(&expr, needle, REG_EXTENDED)) {
```
DanielMolina24, line 146: Can you add a small comment that describes what regcomp is doing

tests/test.c
```diff
@@ -138,10 +138,32 @@ void end_test() {
...
+        fprintf(stderr, "Unable to build regex expression");
+        return NULL;
+    }
+    if (regexec(&expr, haystack, 1, &tracker, REG_EXTENDED)) {
```
DanielMolina24, line 150: Same thing as comment above. A small one line is fine

DanielMolina24 (changes requested):

levincent06 (changes requested):
I've left a comment on using `#define TRUE` and `#define FALSE` (which is a convention we've been using) over defining a global constant variable!

tests/test.c
```diff
@@ -138,10 +138,32 @@ void end_test() {
...
+        return NULL;
+    }
+    return haystack + tracker.rm_so;
+}
+
 // Verifies that expected_output is somewhere in the output
-void in_output(char* expected_output) {
+void in_output(char* expected_output, const uint8_t USE_REGEX) {
+    char *(*checker)(const char *, const char *);
```
levincent06, line 158: Daniel and I were thinking that while using a function pointer here is perfectly valid, it may not be as readable for the general reader! I suggest using something like
```
// Do check
good_check = NULL
if (use_regex)
    good_check = rstrstr(...)
else
    good_check = strstr(...)
// Output result
if (good_check != NULL)
....
```

levincent06 (changes requested):

tests/test.c
DanielMolina24: Yea I agree. The suggested code might make things more readable.

0neDividedbyZer0:
Code should now be updated to be more in style

DanielMolina24 (approved):

levincent06 (approved):
Looks good to me!

levincent06:
Closes #78

PR #128, opened 2020-10-25
[BUILD] Added Update script that checks for runtime package from Dawn

AVSurfer123:


scripts/update.sh
```diff
@@ -0,0 +1,27 @@
...
+done
+
+echo 'IMPORTANT: Runtime Rebooting for Update' >> /tmp/log-fifo
+sleep 2
+
+SERVICES="executor dev_handler net_handler shm_stop shm_start"
+
+cd systemd && sudo systemctl stop $SERVICES
+
+unzip -o /tmp/runtime.zip -d ~/
```
benliao1, line 24: The only question I have here is like did you test to make sure this overwrites the existing runtime folder with no problems (does it ask for permission to overwrite the existing directory?) and on reboot it rebuilds Runtime? Like where/when is it rebuilding Runtime to get the updated executables?

benliao1 (commented):
Just have one question about rebuilding runtime with this~~

scripts/update.sh
AVSurfer123: yes it overwrites the existing folder. it doesn't need to rebuild since the zip contains the built binaries

benliao1: @AVSurfer123  wait how does the zip contain the built binaries? bc the .zip from the repo doesn't contain built binaries right? Unless Dawn scp's the .zip, unzips it, and then builds it?

benliao1: sorry for the late reply

AVSurfer123: oh, is the zip going to be the one directly from the github releases? I thought we could just host a zip of our working pi on the website for them to download. usually source code would be better, but every pi should be the exact same OS image here.

benliao1: I'm not really sure, but I'm pretty sure we're not gonna post the entire raspi disk image on the website...

AVSurfer123: I'm not saying to post the raspi disk image. I said its the ZIP of the runtime folder on a working raspi

benliao1 (approved):
Looks good! 👍

PR #131, opened 2020-11-05
[MISC] Add null checks for malloc

### LINKED ISSUE #121: Add a NULL pointer check after every malloc/calloc
This should be done just so in the rare case that we run out memory, we will log it to our Runtime logger to help debug why our program crashed.

levincent06:
- If null, write a FATAL log and exit(1)
- Didn't add checks for Arduino files and protobuf_test files

Closes #121


benliao1 (approved):
Looks good 👍

DanielMolina24 (approved):

Issue #94, opened 2020-08-05
[LOWCAR] Add a invert_direction parameter to motor controllers


benliao1:
Students should be able to choose which way is forward for their motors with this parameter (a boolean). We'll need to test it to make sure that going forwards and backwards, inverted and non-inverted, result in correct PID control on both motors on the KoalaBear (with correct encoder counts).

PR #140, opened 2021-01-02
[LOWCAR] KoalaBear upgrades

### LINKED ISSUE #92: [LOWCAR] Get Encoder and PID to work on KoalaBears
Encoders and PID control have never worked on PiE motor controllers since probably the GrizzlyBear? Maybe even before then.....

Once the new motors arrive, need to hook them up to the KoalaBears and get them to work.

### LINKED ISSUE #139: [LOWCAR] Test UID for flashed lowcar devices change appear differently within net_handler and dev_handler
When a lowcar device is flashed using the runtime flash script, there is an option to set a test UID instead of a random UID. The test UID is 0x0123456789ABCDEF. However, it seems that when connecting to dev_handler or printing on net_handler_cli shows the UID to be 0x23456789ABCDEF0D. There could be an error on how the UID is being flashed or the way dev_handler could be processing the acknowledgment from the device.

Steps to reproduce:
1) Flash a lowcar device using "runtime flash -t KoalaBear"
2) Start up shared memory and dev_handler
3) Connect flashed lowcar device to dev_handler and observe UID of connected device
4) Print in shared memory to also observe the UID

benliao1:
Implemented and tested a number of KoalaBear features and upgrades:
- KoalaBear encoders work and update in shared memory (#92)
- PID control working on KoalaBears with default values for KP, KI, and KD (#92)
- KoalaBear uses PID controllers by default
- Implemented invert direction parameter which allows students to specify whether clockwise or counterclockwise rotation of a given motor shaft is considered to be "positive duty cycle" (#94)
- Implemented a maximum duty cycle cap; if the cap is, for example, 0.5, then a student input of 0.5 will actually run the motor at 0.5 * 0.5 = 0.25 duty cycle, meaning the fastest the motor will ever go is 0.5 duty cycle
- Removed the `CURRENT_A` and `CURRENT_B` parameters for KoalaBears; these are deprecated parameters from YogiBears that helped implement a software "kill switch" when motors drew too much current. PolarBears and KoalaBears both include this functionality built into the hardware, so there is no need for any software to take care of this any more.

Other issues that this includes:
- Added a few extra checks and log messages in `logger.c` and `tcp_conn.c` to try to catch some memory leaks and other miscellaneous currently un-reproducible bugs
- There was a mistake in `Messenger.cpp` that said that the number of bytes in the device type field of a Lowcar message was 2; it should be 1 (#139)

Does not include:
- Acceleration (although have left a comment to show where I think this will go)
- Fine-tuned PID coefficients (need a lot longer to get really good values for this)

Closes #92 , #94 , #139


lowcar/devices/KoalaBear/PID.cpp
```diff
@@ -1,5 +1,7 @@
 #include "PID.h"
 
+#define MAX_TPS 2700.0   // maximum encoder ticks per second that a motor can go (determined empirically)
```
levincent06, line 3: In which cases would we need to change this? When we get new encoders? If this needs to be changed, how can someone measure the correct value?

lowcar/devices/KoalaBear/KoalaBear.h
```diff
@@ -35,12 +35,12 @@ class KoalaBear : public Device {
...
-    float target_speed_a, target_speed_b;  // target speeds of motors
-    float deadband_a, deadband_b;          // deadbands of motors
-    uint8_t pid_enabled_a, pid_enabled_b;  // whether or not motors are enabled
-    volatile uint32_t enc_a, enc_b;        // encoder values of motors
-    PID *pid_a, *pid_b;                    // PID controllers for motors
-    LEDKoala* led;                         // for controlling the KoalaBear LED
+    float target_speed_a, target_speed_b;           // target speeds of motors
+    float deadband_a, deadband_b;                   // deadbands of motors
+    uint8_t invert_a, invert_b;                     // whether or not motors are going in default direction
+    uint8_t manual_enabled_a, manual_enabled_b;     // whether or not PID on motors is enabled
```
levincent06, line 41: These comments might need to be updated
I like the wording used in `KoalaBear.cpp`:
```
// TRUE if the respective motor should rotate in opposite direction of default; FALSE for default direction
// TRUE if using manual drive mode; FALSE if using PID control (default)
```

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -189,26 +224,37 @@ size_t KoalaBear::device_write(uint8_t param, uint8_t* data_buf) {
 
         // Params for Motor B
         case DUTY_CYCLE_B:
-            this->pid_enabled_b = FALSE;  //remove later for PID functionality
             this->target_speed_b = ((float*) data_buf)[0];
```
levincent06, line 227: We briefly touched on this, but do you think it may be confusing that writing to param `duty_cycle` writes to the `target_speed` field? I can see that keeping it as `duty_cycle` is nice because that's what it's always been but with PID control (and soon acceleration shenanigans), the parameter name may not be as accurate.

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -241,21 +287,17 @@ void KoalaBear::device_enable() {
...
     this->target_speed_b = 0.0;
-    this->pid_enabled_a = FALSE;
-    this->pid_enabled_b = FALSE;
+    this->manual_enabled_a = FALSE;
+    this->manual_enabled_b = FALSE;
 }
 
 void KoalaBear::device_actions() {
-    float target_A, target_B;
+    float command_a, command_b, actual_target_a, actual_target_b;
```
levincent06, line 300: So if I'm understanding correctly,
1. Student writes to `duty_cycle` with a value in [-1.0, 1.0]
2. `Device::device_write()` writes that exact value to `this->target_speed`
3. `actual_target_speed` is `this->target_speed` adjusted for `MAX_DUTY_CYCLE`, acceleration, and invertedness
4. `command` is `actual_target_speed` adjusted with PID control (if applicable)
5. `command` is the _actual_ duty cycle written to the pin

I was thinking it might(?) be clearer if the names were:
- param `duty_cycle` --> `target_speed`
- `this->target_speed` -- keep the same (student writing to `target_speed` writes to `target_speed`)
- `this->actual_target_speed` -- keep the same or `this->adjusted_target_speed`
- command --> `duty_cycle`

Are we still considering changing the "speed" units to encoder ticks per second in the future?

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -189,26 +224,37 @@ size_t KoalaBear::device_write(uint8_t param, uint8_t* data_buf) {
...
+	    case INVERT_B:
+			this->invert_b = data_buf[0];
+			return sizeof(uint8_t);
+        case MANUAL_ENABLED_B:
+            this->manual_enabled_b = data_buf[0];
             return sizeof(uint8_t);
         case PID_KP_B:
-            this->pid_b->set_coefficients(((double*) data_buf)[0], this->pid_b->get_ki(), this->pid_b->get_kd());
+			// ignore executor setting all coefficients to 0 when going into idle
+			if (((double*) data_buf)[0] != 0.0) {
```
levincent06, line 243: This is the same as saying "ignore any 'zero' writes to the PID coefficients?"

logger/logger.c
```diff
@@ -283,14 +283,21 @@ void log_printf(log_level_t level, char* format, ...) {
...
+    static int ret;                      // holds return values for lib functions
     va_list args;                        // this holds the variable-length argument list
 
     // lock the mutex around all output functions
     pthread_mutex_lock(&log_mutex);
 
     // expands the format string into msg
     va_start(args, format);
-    vsprintf(msg, format, args);
+    if ((ret = vsnprintf(msg, sizeof(msg), format, args)) != 0) {
```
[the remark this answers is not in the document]
benliao1: Yes, makes sense, fixed!

lowcar/devices/KoalaBear/PID.cpp
benliao1: Yeah, when we get new encoders / new motors or potentially with the new PDB or power distribution system.. all I did was took my four motors, ran them at 1.0 duty cycle for 1 second 5 times each in both directions and recorded how far the encoders ticked, then took the average of all of those readings and it came out to like 2697 ish so I rounded it up to 2700 for simplicity.

lowcar/devices/KoalaBear/KoalaBear.h
benliao1: Yes, fixed--although the `manual_enabled_a` stuff might need to be updated (see later comment)

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: Yeah, I like this, I wanted to change it as well but wasn't sure if other people would agree with me on this. Since you also want to change the name, I'll change it!

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: Yes, what you've got there is exactly right and I'll change it now. The "speed" units I think we're gonna keep it to -1.0 to 1.0 still, since I think it's a bit more intuitive for students than the somewhat arbitrary range -2700.0 to 2700.0 "ticks per second" (what's a tick? why can't it go higher than 2700 tps? etc.).

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: Ok so this was my (maybe) temporary solution to a problem I was struggling to solve. So Ashwin made it so that `executor` writes 0 (`int`s), 0.0 (`float`s), and False (`bool`s) to all writeable parameters when the run mode is switched to `IDLE` or when `executor` crashes. This is almost always the right thing to do, but it's definitely *not* the right thing to do for the PID coefficients (and one could argue it's not the right thing to do for the encoder values either). So without this here, currently what happens is that when you plug in the KoalaBear and the device enables, the PID coefficients are correct and set to the default values, and when you put the robot in auto or teleop, it works. But then when you switch it to `IDLE` all the motors stop working because the PID is by default enabled and all the coefficients were set to 0.

Putting these checks here solves the problem but it also makes it so that you can't write a 0 to, say KI--the device will literally just reject it. That's actually something a student might want to do (just have KP and no KI or KD), so it could be a problem. However, it's not something they're *likely* to do... if they're trying to set coefficients to zero you're supposed to just disable PID, but this solution is not the best.

Additionally, the reason why I had to change the parameter `pid_enabled_a` to `manual_enabled_a` is because of the very same issue. If I kept it to `pid_enabled_a`, `executor` would simply disable it immediately when going to `IDLE`, and from the KoalaBear's point of view, there's no way of telling whether that disable was from `executor` switching run modes or from the student actually disabling PID. I'd prefer to keep it as `pid_enabled_a` and not `manual_enabled_a` because `pid_enabled_a` sounds a bit more "correct"--like you're enabling a feature and then disabling it for more control.... does that make sense?

Anyways, I'm not sure what the best way to fix this problem is.... what do you think?

levincent06: If I'm getting this straight, executor will "zero" out write-able parameters on `IDLE` or when executor crashes. This is the right move for most cases, except for when we have non-zero default values? This is the first time we're encountering this case

levincent06: Something I'm thinking about is that in the lowcar code, we can define default values if necessary and executor will send a new `DEVICE_RESET` message (in addition to our other lowcar message types). When a lowcar device receives this message, it can write to parameters/pins on its own as appropriate. This would mean defining `Device::device_reset()`. Maybe this has to do with how we used to have a `Device::device_disable()`?? Unsure whether "disable" or "reset" is the more appropriate wording here.

benliao1: That's right, what you said in the first comment is exactly what's happening.
Do you think it's worth it to add a whole new lowcar message for this purpose? You're right, I think we can perhaps use this opportunity to reimagine the purpose of `device_disable()` in `lowcar`.... let's talk about this and I'll write up the results of the discussion in a later reply.

lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -28,24 +28,33 @@
...
+//******************************* KOALABEAR CONSTANTS AND PARAMS ****************************//
+
+// ----------------------> CHANGE THIS TO IMPLEMENT MAX SPEED FOR KARTS <----------------------
+#define MAX_DUTY_CYCLE 0.5 // maximum duty cycle to cap how fast the motors can go: range (0, 1] 
+// --------------------------------------------------------------------------------------------
+
+// default values for PID controllers
+#define KP_DEFAULT 0.035
+#define KI_DEFAULT 0.005
+#define KD_DEFAULT 0.0
```
DanielMolina24, line 40: Would it be good to add a small comment to each of these variables or description above on what each of these PID values represent? Or maybe a mention on how the wiki covers what they are(even though its not super in depth there and might need to be expanded on)? 
Also out of curiosity why are these the default values? I know KD has to be zero since there has obviously been no movement whatsoever so you cant calculate the change in error, but what about the other two?

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: Solution:
Change `executor` to only reset the params that it needs to, i.e. only the `velocity_a` and `velocity_b` params on the KoalaBear as of now. (Much simpler than expected! 👍 )

lowcar/devices/KoalaBear/KoalaBear.cpp
benliao1: Yes, I think I'm going to add a comment saying that the PID controller is talked about in the wiki... as far KD being zero actually that is a little unusual, but for our motors it seems that it's a little too responsive to it so I set to 0 for the time being!

DanielMolina24 (approved):

PR #147, opened 2021-01-05
[BUILD] Fixed all compiler warnings

AVSurfer123:
Fixed all the compiler warnings. Couldn't figure out a nice/easy way to have `uint64_t` have the correct print formatter on both Windows and Linux so I chose to stick with the Linux default of `%lu` for `uint64_t`, where before we used the Windows default of `%llu`. There is one way of doing it using `inttypes.h` but its not very elegant. 

Also, I changed the `./runtime build` command to pass any arguments so now to force a rebuild, you can just run `./runtime build -B`.

### DIFF
--- a/executor/executor.c
+++ b/executor/executor.c
@@ -395,7 +395,7 @@ static void run_challenges() {
             outputs.payload[i] = "Fatal error";
             continue;
         }
-        int ret_len;
+        long ret_len;
         const char* c_ret = PyUnicode_AsUTF8AndSize(ret_string, &ret_len);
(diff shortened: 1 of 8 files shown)


AVSurfer123:
Nvm I'm dumb. The main difference of `%llu` vs `%lu` is on 32bit vs 64bit systems so the 32 bit Pi will always have different definition of a long compared to our 64 bit computers. This PR isn't necessary as the default is correct and will work on all systems despite warnings. Using `%lu` on the Pi will break things, which the test case found.

PR #153, opened 2021-01-09
[TEST] Fix net_handler_cli bug with device data

levincent06:
- Net Handler CLI
  - Properly malloc and populate device names for subs
  - Prompt "done?" only after at least one subscription
  - Make pressing enter as a response to default actions
- Shared Memory
  - Add/clarify comments in shm_start and shm_wrapper

### DIFF
--- a/shm_wrapper/shm_wrapper.c
+++ b/shm_wrapper/shm_wrapper.c
@@ -779,7 +779,7 @@ int gamepad_read(uint32_t* pressed_buttons, float joystick_vals[4]) {
         for (int i = 0; i < 4; i++) {
             joystick_vals[i] = 0;
         }
-        log_printf(DEBUG, "tried to read, but no gamepad connected");
+        log_printf(ERROR, "tried to read, but no gamepad connected");
         my_sem_post(rd_sem, "robot_desc_mutex");
         return -1;
     }
--- a/shm_wrapper/shm_wrapper.c
+++ b/shm_wrapper/shm_wrapper.c
@@ -659,7 +659,7 @@ int place_sub_request(uint64_t dev_uid, process_t process, uint32_t params_to_su
         curr_sub_map[dev_ix + 1] = params_to_sub;
         curr_sub_map[0] |= (1 << dev_ix);  // turn on corresponding bit
     }
-
+    
     // release sub_map_sem
     my_sem_post(sub_map_sem, "sub_map_sem");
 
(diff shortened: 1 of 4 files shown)


shm_wrapper/shm_wrapper.c
```diff
@@ -775,7 +775,11 @@ int gamepad_read(uint32_t* pressed_buttons, float joystick_vals[4]) {
 
     // if no gamepad connected, then release rd_sem and return
     if (rd_shm_ptr->fields[GAMEPAD] == DISCONNECTED) {
-        log_printf(ERROR, "tried to read, but no gamepad connected");
+        *pressed_buttons = 0;
+        for (int i = 0; i < 4; i++) {
+            joystick_vals[i] = 0;
+        }
+        log_printf(DEBUG, "tried to read, but no gamepad connected");
```
AVSurfer123, line 782: It is an error on purpose, so the student knows they are running faulty code. As of now, this is only called in the student API. Right now, it looks like it will print 2 error statements, one from the logger and one from a Python exception. Is that why you wanted to make it debug?

tests/cli/net_handler_cli.c
```diff
@@ -299,42 +292,51 @@ void prompt_device_data() {
                 return;
             }
 
+            // Assign type and make sure it's an integer
+            dev_type = (uint8_t) strtol(nextcmd, NULL, 10);
             errno = 0;
```
AVSurfer123, line 297: `errno` needs to be set to 0 before you call `strtol`.

AVSurfer123 (changes requested):

tests/cli/net_handler_cli.c
levincent06: You're right! That's a mistake--I'll make that change (with a comment explaining why).

shm_wrapper/shm_wrapper.c
levincent06: I can switch it back to `ERROR`. How does the student know that their code is faulty if this log appears? This means either that 1) they don't have a gamepad connected in `TELEOP` or 2) they're trying to access gamepad data in `AUTO`?

The main reason I changed it was because it was giving me issues while working on the shm UI. The UI polls for the gamepad state multiple times per second. I tested each component of the UI individually, many of which didn't need a gamepad connected. As a result, my logs were being overwhelmed with the `ERROR` message, so I couldn't use other logs to debug my work. My temporary workaround was to simply comment out the log.

If any user wants to use the shm UI, they'll want to hide this log message so that they can see other logs. What do you recommend I should do here?

AVSurfer123 (approved):

benliao1 (approved):
I like the flow of this logic, it's definitely an improvement over my hastily-written CLI logic. I see that you prefer the separation of assignment and checking for errors into two lines--that works too. I tend to prefer assigning and checking for errors together, but maybe that's because I've read K&R too many times and that's what they do and so it makes sense to me. Thanks for adding clarifying comments too!

AVSurfer123:
I also prefer separating assignment as it makes it clearer. One line should perform one function

PR #154, opened 2021-01-09
Shared Memory UI

### LINKED ISSUE #127: [SHM] Htop-style live-view debugging tool
I'm imagining some graphical representation of shared memory in the terminal that updates continuously, like htop, so that when we're debugging shared memory, we can see new devices coming on and off in real time as new blocks coming in and out of view in the terminal, and each device has a nicely formatted list of all parameters.

Reach out if you have any questions about this!

levincent06:
The UI uses ncurses, a third-party dependency to provide real-time visualization of data flowing through shared memory.
ncurses can be installed manually with `sudo apt-get -y install libncurses5-dev libncursesw5-dev`.  (The current Docker image on master already has this dependency.)

This UI replaces shm_cli, though it still uses shm_client.

The UI displays:
- Robot Description
  - Connected-ness of Dawn, Shepherd, and the gamepad
  - Run Mode
  - Start Position
- Gamepad State
  - Joystick values and pressed buttons
- Device-specific information
  - Shared memory index, device name/type/year/uid
  - For each parameter:
     - Index, name, command stream, data stream
     - Whether it is being subscribed to by Net Handler and by Executor 

The UI displays information about only one specific device at a time. The user can choose which device to view by using the left and right arrow keys.

Closes #127 

EDIT: Also displays custom data block in shared memory as the `MAX_DEVICES`-th device.


benliao1:
Yeah the code looks really good and well documented:
- Check gamepad to make sure it's connected before polling for the gamepad data
- Delete the printing functions in `shm_wrapper.c` and `shm_wrapper.h`

DanielMolina24 (approved):

tests/cli/shm_ui.c
```diff
@@ -0,0 +1,471 @@
...
+    // Start at line after the header
+    int line = 2; // Note that this is just a row number; Must use wmove() to actually move the cursor to this line
+    wmove(GAMEPAD_WIN, line, 0);
+    wclrtoeol(GAMEPAD_WIN); // Clear "No gamepad connected"
+
+    // Read gamepad state if gamepad is connected
+    uint32_t pressed_buttons = 0;
+    float joystick_vals[4] = {0};
+    int gamepad_connected = (robot_desc_read(GAMEPAD) == CONNECTED) ? 1 : 0;
+    if (gamepad_connected) {
```
AVSurfer123, line 166: No need to have this int `gamepad_connected` and the ternary expression. Just directly compare the robot desc to `CONNECTED` in the if statement.

tests/cli/shm_ui.c
```diff
@@ -0,0 +1,471 @@
...
+    } else if (DEVICE_WIN_IS_BLANK) { // Window was previously clear (i.e Switching from no device to a connected device)
+        // Put back title and box
+        mvwprintw(DEVICE_WIN, 1, 1, DEVICE_WIN_HEADER);
+        mvwprintw(DEVICE_WIN, DEVICE_HEIGHT - 2, 1, NOTE);
+        box(DEVICE_WIN, 0, 0);
+        wrefresh(DEVICE_WIN);
+        // Note that we will display data, so DEVICE_WIN will no longer be "empty"
+        DEVICE_WIN_IS_BLANK = 0;
+    } else { 
+        // Switching from one connected device to another
```
AVSurfer123, line 244: Why have this empty `else` block?

tests/cli/shm_ui.c
```diff
@@ -0,0 +1,471 @@
...
+
+    // Print current device's identifiers
+    device_t* device = NULL;
+    wclrtoeol(DEVICE_WIN); // Clear previous string
+    if (show_custom_data) {
+        mvwprintw(DEVICE_WIN, line++, 1, "Custom Data");
+    } else {
+        device = get_device(dev_ids[shm_idx].type);
+        if (device == NULL) { // This should never happen if the handling above is correct
+            log_printf(DEBUG, "device == NULL");
```
AVSurfer123, line 259: if this should never happen, make the log_printf an `ERROR`

tests/cli/shm_ui.c
```diff
@@ -0,0 +1,471 @@
...
+
+    // Init variables to hold custom data information
+    char custom_param_names[UCHAR_MAX][64];
+    param_type_t custom_param_types[UCHAR_MAX];
+    param_val_t custom_param_values[UCHAR_MAX]; // We're assuming here that the number of logged params is at most MAX_PARAMS
+
+    // Get shm subscriptions
+    if (show_custom_data) {
+        log_data_read(&num_params, custom_param_names, custom_param_types, custom_param_values);
+        // log_printf(DEBUG, "Showing %d custom params", num_params);
```
AVSurfer123, line 285: remove debug statement

tests/cli/shm_ui.c
```diff
@@ -0,0 +1,471 @@
...
+        }
+
+        // Display values according to the parameter's type
+        param_type_t param_type = (show_custom_data) ? custom_param_types[i] : device->params[i].type;
+        switch (param_type) {
+            case INT:
+                if (display_cmd_val) {
+                    mvwprintw(DEVICE_WIN, line, COMMAND_VAL_COL, "%d", command_vals[i].p_i);
+                }
+                mvwprintw(DEVICE_WIN, line, DATA_VAL_COL, "%d", (show_custom_data) ? custom_param_values[i].p_i : data_vals[i].p_i);
```
AVSurfer123, line 342: One way to reduce code is to move the ternary condition before the switch to get the `param_val_t`, similarly to what you did for the param type. Then you can just use that one struct in each of the 3 cases

tests/cli/shm_ui.c
```diff
@@ -0,0 +1,471 @@
...
+}
+
+int main(int argc, char** argv) {
+    // signal handler to clean up shm
+    signal(SIGINT, sigint_handler);
+    logger_init(TEST);
+
+    // Start shm
+    start_shm();
+    sleep(1); // ALlow shm to initialize
```
AVSurfer123, line 405: typo

README.md
```diff
@@ -89,6 +89,11 @@ Then, extract the files with
 
 You might need to install `wget` and `tar` with `sudo apt-get -y install wget tar`.
 
+### Installing ncurses for Shared Memory UI
+The Shared Memory UI allows the user to have a real-time view of data flowing in and out of shared memory.
+This is analogous to the "shm_cli" among dev_handler_cli, executor_cli, and net_handler_cli.<br />
+`sudo apt-get -y install libncurses5-dev libncursesw5-dev`
```
AVSurfer123, line 95: Add that this is for development only, and not for production.

AVSurfer123 (changes requested):

tests/cli/shm_ui.c
levincent06: It's there only for clarity to understand the logic if someone else were to make changes to the UI in the future.

tests/cli/shm_ui.c
levincent06: Makes sense. I had the extra variable because I used it later in this function multiple times, but I removed them.

levincent06:
The above force-push was a result of rebasing off of master following the recent PRs.

benliao1 (approved):
Looks good; the issues mentioned in the previous comment will be addressed later so this should be merged in ASAP!

tests/cli/shm_ui.c
```diff
@@ -166,38 +173,53 @@ void display_gamepad_state(char** joystick_names, char** button_names) {
...
-    line += 2;
+    line++;
     wmove(GAMEPAD_WIN, line, 0);
 
     // Print pressed buttons
-    mvwprintw(GAMEPAD_WIN, line++, 1, "Pressed Buttons:");
-    wmove(GAMEPAD_WIN, line, 0);
+    wclrtoeol(GAMEPAD_WIN);
+    if (gamepad_connected) {
+        mvwprintw(GAMEPAD_WIN, line, 1, "Pressed Buttons:");
```
AVSurfer123, line 208: is it necessary to go through all the joystick and gamepad logic when the gamepad isn't connected? why can't you just return after the first gamepad check above?

levincent06: I wish it was that easy :P The reason is because at each frame, ncurses draws over the last "snapshot". So lets say all the gamepad stuff was displayed, and now we don't want to display them anymore. We have to explicitly say to clear all that data with whitespace. There are two options here:
1) Clear the entire gamepad window, and draw in the header, box, etc back again
2) Do what I'm doing rn: print iff gamepad is connected

I opted for option 1 for `DEVICE_WIN`. Option 1 has a lot of book-keeping (see that big logic block at the beginning of `display_device_win()`. It's nice to clear a ton of characters at once, but the logic is kinda messy imo and requires global variables.

Option 2 is a bit easier to follow and doesn't require having a global variable. I think this is nicer for smaller windows like `GAMEPAD_WIN`

AVSurfer123: ok i see makes sense that you have to overwrite with whitespace

AVSurfer123 (approved):

Issue #155, opened 2021-01-10
[SHM] Remove print utility functions


levincent06:
Shared memory wrapper currently has functions to print blocks of data, including gamepad state, run mode, etc. This was used extensively in the test cases, but the test API has been rewritten to compare values directly with shared memory data (no more output string comparison). These print functions were also used in the shm cli, which will be replaced by the shm UI. As a result, these print functions are no longer useful.

However, they are currently used in a couple files like `integration/tc_71_1` in `shm_wrapper/shm_wrapper_test1.c`. These should be rewritten to not use the print functions.

PR #157, opened 2021-01-10
Added a new Keyboard class to the Student API

### LINKED ISSUE #149: [STUDENTAPI] Add a Keyboard class to the Student API which access shared memory

### LINKED ISSUE #150: [SHM] Add a new SHM block to hold keyboard data
The new block will hold keyboard data from Dawn

### LINKED ISSUE #151: [NET_HANDLER] Add a new keyboard proto and handle it in udp_conn
The new proto should just be a 64bit integer which will be used as a bitmask for certain keys. This will need to be handled in `udp_conn.c` very similarly to the gamepad data.

AVSurfer123:
This requires changing many things.

1. Need to restructure the SHM block for gamepad to be for all inputs. Currently it only has 2 inputs, `GAMEPAD` and `KEYBOARD`. Also added another field to the robot description SHM block
2. Need to add a Keyboard class in Cython `studentapi.pyx` which accesses shared memory, and need to insert the class into the namespace in `executor_init`.
3. Need to restructure the `gamepad.proto` to `input.proto` which handles inputs from the gamepad and keyboard together. This will require changing the Dawn client, which I will do next. 
4. Need to rework `udp_conn.c` to use the new `input.proto` and write to the appropriate blocks in SHM
5. Need to rewrite how to send inputs in the `net_handler_client` and the integration tests.
6. Need to add integration tests, 1 is already one in this PR.

TODO:
1. Need to rewrite the `net_handler_cli` to allow sending keyboard inputs.
2. Add more tests

Closes #149, closes #150, closes #151


shm_wrapper/shm_wrapper.h
```diff
@@ -45,17 +49,26 @@ typedef struct {
...
-    float joysticks[4];  // array to hold joystick positions
-} gamepad_shm_t;
+    uint64_t buttons;           // bitmap for which buttons are pressed
+    float joysticks[4];         // array to hold joystick positions, only for gamepad
+    robot_desc_field_t source;  // which hardware device the command came from, either GAMEPAD or KEYBOARD
+} input_t;
+
+// shared memory for gamepad and keyboard inputs
+typedef struct {
+    input_t inputs[2];  // Index 0 is for GAMEPAD, index 1 is for KEYBOARD, like in the enum. can be modified in the future
```
levincent06, line 62: I mentioned this in another comment but you can use the `#define` here for the 2

shm_wrapper/shm_wrapper.h
```diff
@@ -282,11 +295,12 @@ void robot_desc_write(robot_desc_field_t field, robot_desc_val_t val);
  * Arguments:
  *    pressed_buttons: pointer to 32-bit bitmap to which the current button bitmap state will be read into
```
levincent06, line 296: Update 32-bit to 64-bit!

shm_wrapper/shm_wrapper.h
```diff
@@ -295,11 +309,12 @@ int gamepad_read(uint32_t* pressed_buttons, float joystick_vals[4]);
  *    pressed_buttons: a 32-bit bitmap that corresponds to which buttons are currently pressed
  *        (only the first 17 bits used, since there are 17 buttons)
```
levincent06, line 310: Update 32-bit to 64-bit! And update 17 too

scripts/test.sh
```diff
@@ -66,6 +66,13 @@ function run_tests {
 # installs function clean_up as SIGINT handler
 trap 'sigint_handler' INT
 
+getopts "e" opt
+if [ $opt != "e" ]; then
+	# Make errors exit bash script
+	set -e
+fi
+shift $(($OPTIND - 1))
+
```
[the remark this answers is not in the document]
levincent06: Also, it doesn't seem like `logger.config` is put back to its original state if a test case fails and script exits early.

shm_wrapper/shm_wrapper.c
```diff
@@ -367,30 +370,37 @@ void print_robot_desc() {
...
-    printf("Current Gamepad State:\n");
-    if (gamepad_read(&buttons, joysticks) == -1) {
-        printf("\tNo gamepad currently connected\n");
-    } else {
-        printf("\tPushed Buttons:\n");
-        // only print pushed buttons (so we don't print out many lines of output each time we all this function)
-        for (int i = 0; i < NUM_GAMEPAD_BUTTONS; i++) {
-            if (buttons & (1 << i)) {
-                printf("\t\t%s\n", button_names[i]);
+    for (int i = 0; i < 2; i++) {
```
levincent06, line 380: Comment about `#define` and 2

shm_wrapper/shm_wrapper.c
```diff
@@ -784,26 +799,32 @@ int gamepad_read(uint32_t* pressed_buttons, float joystick_vals[4]) {
...
     // wait on gp_sem
-    my_sem_wait(gp_sem, "gamepad_mutex");
+    my_sem_wait(input_sem, "input_mutex");
 
-    *pressed_buttons = gp_shm_ptr->buttons;
+    int index = source - GAMEPAD;  // Assumes KEYBOARD is after GAMEPAD in robot_desc_field_t enum
+    *pressed_buttons = input_shm_ptr->inputs[index].buttons;
     for (int i = 0; i < 4; i++) {
-        joystick_vals[i] = gp_shm_ptr->joysticks[i];
+        joystick_vals[i] = input_shm_ptr->inputs[index].joysticks[i];
```
levincent06, line 807: We read the joystick values iff the source is `GAMEPAD`, right?

shm_wrapper/shm_wrapper.c
```diff
@@ -812,15 +833,16 @@ int gamepad_write(uint32_t pressed_buttons, float joystick_vals[4]) {
...
     // wait on gp_sem
-    my_sem_wait(gp_sem, "gamepad_mutex");
+    my_sem_wait(input_sem, "input_mutex");
 
-    gp_shm_ptr->buttons = pressed_buttons;
+    int index = source - GAMEPAD;  // Assumes KEYBOARD is after GAMEPAD in robot_desc_field_t enum
+    input_shm_ptr->inputs[index].buttons = pressed_buttons;
     for (int i = 0; i < 4; i++) {
-        gp_shm_ptr->joysticks[i] = joystick_vals[i];
+        input_shm_ptr->inputs[index].joysticks[i] = joystick_vals[i];
```
levincent06, line 841: Same comment as above regarding joysticks being valid for only `GAMEPAD`

shm_wrapper/shm_start.c
```diff
@@ -54,36 +54,36 @@ int main() {
...
-        log_printf(FATAL, "mmap: %s", strerror(errno));
+        log_printf(FATAL, "mmap devices: %s", strerror(errno));
         exit(1);
     }
     if (close(fd_shm) == -1) {
-        log_printf(ERROR, "close: %s", strerror(errno));
+        log_printf(ERROR, "close devices: %s", strerror(errno));
     }
 
     // create gamepad shm block
```
levincent06, line 72: Update to say inputs shm block

levincent06 (changes requested):

tests/integration/tc_150_1.c
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Tests the Keyboard input source through net_handler, SHM, and executor. 
+ */
+
+int main() {
+    start_test("keyboard_input", "keyboard_input", "", NO_REGEX);
+
+    connect_virtual_device("SimpleTestDevice", 20);
+
```
levincent06, line 11: I would suggest adding a `sleep(1)` after calling this function, as it can take a bit for the device to connect. From adding a couple print statements, I found that the device doesn't connect until after `TELEOP` started. (I posted a screenshot in the Slack thread, nearly 10 lines later)

levincent06 (changes requested):

net_handler/udp_conn.c
```diff
@@ -197,36 +197,39 @@ static void* send_device_data(void* args) {
...
-
-        robot_desc_write(GAMEPAD, gp_state->connected ? CONNECTED : DISCONNECTED);
-        if (gp_state->connected) {
-            if (gp_state->n_axes != 4) {
-                log_printf(ERROR, "update_gamepad_state: Number of joystick axes given is %d which is not 4. Cannot update gamepad state", gp_state->n_axes);
-            } else {
-                gamepad_write(gp_state->buttons, gp_state->axes);
+        for (int i = 0; i < inputs->n_inputs; i++) {
+            Input* input = inputs->inputs[i];
+            robot_desc_field_t source = GAMEPAD + input->source;  // Assumes that order in enum is GAMEPAD, KEYBOARD
```
DanielMolina24, line 222: So `input->source` is an int right? So if the `input->source` is GAMEPAD, the small addition here is just 0(GAMEPAD) + 0(input->source)?

scripts/test.sh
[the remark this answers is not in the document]
DanielMolina24: Yea I agree! Some test cases have similar issues making debugging easier when looking at all the failed ones together!

net_handler/udp_conn.c
DanielMolina24: Ah never mind, just realized youre iterating over inputs which cuases input to change each time

shm_wrapper/shm_wrapper.c
```diff
@@ -837,12 +859,18 @@ int log_data_write(char* key, param_type_t type, param_val_t value) {
...
         log_printf(ERROR, "Maximum number of %d log data keys reached. can't add key %s", UCHAR_MAX, key);
         return -1;
     }
 
+    // 64 is max key size in log data SHM block
+    if (strlen(key) >= 64) {
+        log_printf(ERROR, "Key name %s for log data is longer than 64 characters", key);
+        return -2;
+    }
+
```
DanielMolina24, line 873: Can you explain what a log data key is? I'm aware that the inputs protobuf accounts for 64 different buttons available across the gamepad and keyboard. Is this related to that, such as a student somehow exceeding that limit?

tests/client/net_handler_client.c
```diff
@@ -407,36 +407,45 @@ void send_start_pos(robot_desc_field_t client, robot_desc_val_t pos) {
...
+        input.axes[i] = joystick_vals[i];
     }
-    len = gp_state__get_packed_size(&gp_state);
-    send_buf = malloc(len);
+
+    uint16_t len = user_inputs__get_packed_size(&inputs);
+    uint8_t* send_buf = malloc(len);
     if (send_buf == NULL) {
-        printf("send_gamepad_state: Failed to malloc\n");
+        printf("send_user_input: Failed to malloc buffer\n");
```
DanielMolina24, line 439: Vin addition to Vincent's comment about freeing above, we should also be freeing `inputs.inputs` and `inputs.axes` before returning or exiting in the event of an error right?

shm_wrapper/shm_wrapper.c
```diff
@@ -812,15 +833,16 @@ int gamepad_write(uint32_t pressed_buttons, float joystick_vals[4]) {
     my_sem_post(rd_sem, "robot_desc_mutex");
 
     // wait on gp_sem
-    my_sem_wait(gp_sem, "gamepad_mutex");
+    my_sem_wait(input_sem, "input_mutex");
 
-    gp_shm_ptr->buttons = pressed_buttons;
+    int index = source - GAMEPAD;  // Assumes KEYBOARD is after GAMEPAD in robot_desc_field_t enum
+    input_shm_ptr->inputs[index].buttons = pressed_buttons;
```
benliao1, line 839: I wonder if we can add a check here to make sure that they aren't any bits on at higher positions than the number of buttons on the input source (for example from passing in a button bitmap intended for a keyboard as a gamepad instead--gamepads don't have 47 buttons)

net_handler/protos/input.proto
```diff
@@ -6,8 +6,18 @@ syntax = "proto3";
...
+message Input {
     bool connected = 1;
-    fixed32 buttons = 2;
+    fixed64 buttons = 2;
     repeated float axes = 3;
+    Source source = 4;
+}
+
+message UserInputs {
+    repeated Input inputs = 1;
```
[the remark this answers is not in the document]
AVSurfer123: yes

tests/client/executor_client.h
```diff
@@ -1,10 +1,13 @@
...
-#include <stdlib.h>
-#include <sys/wait.h>
-#include "../../runtime_util/runtime_util.h"
+#include <errno.h>     // for errno
+#include <signal.h>    // for SIGINT
+#include <stdio.h>     // for printf, stderr
+#include <stdlib.h>    // for malloc, free, realloc, exit
+#include <string.h>    // for strcat, strcpy
+#include <sys/wait.h>  // for waitpid, pid_t
+#include <unistd.h>    // for execlp, fork, chdir
```
benliao1, line 10: Not sure if this was autofill from VSCode but I liked it the previous way (although the comments are useful). The reason for the `runtime_util` include is because it's meant to be a place where a bunch of headers that are commonly included are included, and the reader can know like "oh, he included that file, which I know contains such-and-such headers".....

tests/client/net_handler_client.c
```diff
@@ -407,36 +407,45 @@ void send_start_pos(robot_desc_field_t client, robot_desc_val_t pos) {
...
 
     // build the message
-    gp_state.connected = 1;
-    gp_state.buttons = buttons;
-    gp_state.n_axes = 4;
-    gp_state.axes = malloc(sizeof(double) * 4);
-    if (gp_state.axes == NULL) {
-        printf("send_gamepad_state: Failed to malloc\n");
+    inputs.n_inputs = 1;
+    inputs.inputs = malloc(sizeof(Input*) * inputs.n_inputs);
```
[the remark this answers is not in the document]
benliao1: +1 Yes

runtime_util/runtime_util.h
```diff
@@ -74,6 +75,16 @@ typedef enum gp_joysticks {
     JOYSTICK_RIGHT_Y
 } gp_joystick_t;
```
benliao1, line 76: Yes this is really annoying, but could we list out enums for all 47 buttons in the same order they appear in that key list? That way we can refer to the buttons with their names as enums in tests and CLI's, instead of doing things like `(1 << 2);  <-- 'z ` in our code?

tests/logger.config
```diff
@@ -23,7 +23,7 @@ STDOUT_LOG_LEVEL: DEBUG
 FILE_LOG_LEVEL: DEBUG
 
 // Print logs at this level and above to NETWORK, must be in all caps.
-NETWORK_LOG_LEVEL: INFO
+NETWORK_LOG_LEVEL: DEBUG
```
benliao1, line 26: Was this an intended production change? Pretty sure we decided on not sending `DEBUG` logs to Dawn?

tests/test.c
```diff
@@ -429,9 +429,9 @@ void add_unordered_string_output(char* output) {
 // ************************* GAMEPAD CHECK FUNCTIONS ************************ //
 
 // Helper function to print a 32-bitmap to stderr on a new line
-static void print_bitmap(uint32_t bitmap) {
+static void print_bitmap(uint64_t bitmap) {
     uint8_t bit;
-    for (int i = 0; i < 32; i++) {
+    for (int i = 0; i < 64; i++) {
```
benliao1, line 434: Are all bitmaps that we print 64 bits? If so then this is OK. If not, I think we probably want like a second argument to this function that specifies how many bits to print, since this would fail trying to print a 32-bit int right now

benliao1 (changes requested):
Good work! A lot of changes littered all over the codebase and a few questions I have, but overall this should work. Like you said though, there remains work to be done after this is merged in!

tests/client/net_handler_client.c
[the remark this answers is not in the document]
AVSurfer123: Yes, thanks I missed it

net_handler/udp_conn.c
```diff
@@ -197,36 +197,39 @@ static void* send_device_data(void* args) {
 }
 
 
-static void* update_gamepad_state(void* args) {
-    static int size = sizeof(GpState);
+static void* update_inputs(void* args) {
+    static int size = 128;
```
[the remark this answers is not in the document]
AVSurfer123: honestly I'm not sure. it just needs to be big enough to accept all `UserInput` messages from Dawn, but I don't know what that size is so I just gave a large enough value of 128. i can add a comment though

executor/executor.c
```diff
@@ -213,7 +229,7 @@ static uint8_t run_py_function(const char* func_name, struct timespec* timeout,
             //if the time the Python function took was greater than max_time, warn that it's taking too long
             time = (end.tv_sec - start.tv_sec) * 1e9 + (end.tv_nsec - start.tv_nsec);
             if (timeout != NULL && time > max_time) {
-                log_printf(WARN, "Function %s is taking longer than %lu milliseconds, indicating a loop in the code.", func_name, (long) (max_time / 1e6));
+                log_printf(WARN, "Function %s is taking longer than %lu milliseconds, indicating a loop or sleep in the code. This is probably not what you intended.", func_name, (long) (max_time / 1e6));
```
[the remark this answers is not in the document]
AVSurfer123: Yea that's what I was trying to do here. Should I be more specific I guess?

scripts/stop.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
AVSurfer123: Oh yea, forgot to remove my debug statements.

tests/Makefile
```diff
@@ -48,44 +39,35 @@ keyboard_interface: cli/keyboard_interface.c ../net_handler/*.* client/net_handl
...
 
-devices: general_dev simple_dev unresponsive_dev foreign_dev unstable_dev time_dev sound_dev
+VIRTUAL_DEVICE_FILES := client/virtual_devices/virtual_device_util.* ../dev_handler/message.* ../runtime_util/runtime_util.* ../logger/logger.c
 
-VIRTUAL_DEVICE_FILES := client/virtual_devices/virtual_device_util.o ../dev_handler/message.o ../runtime_util/runtime_util.o ../logger/logger.o
+VIRTUAL_DEVICES := $(shell ls client/virtual_devices/*Device.c | awk -F'.' '{ print $$1 }')
 
-general_dev: client/virtual_devices/GeneralTestDevice
+devices: $(VIRTUAL_DEVICES)
+	echo "virtual devices made" > devices
```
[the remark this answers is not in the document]
AVSurfer123: `gcc` will always repeat what it does which is why you see that. but this does work as intended as you can only do `make devices` once. This was the hack I mentioned earlier because we create a file called `devices`.

tests/Makefile
```diff
@@ -48,44 +39,35 @@ keyboard_interface: cli/keyboard_interface.c ../net_handler/*.* client/net_handl
 
 ################################ VIRTUAL DEVICES ###############################
 
-devices: general_dev simple_dev unresponsive_dev foreign_dev unstable_dev time_dev sound_dev
+VIRTUAL_DEVICE_FILES := client/virtual_devices/virtual_device_util.* ../dev_handler/message.* ../runtime_util/runtime_util.* ../logger/logger.c
 
-VIRTUAL_DEVICE_FILES := client/virtual_devices/virtual_device_util.o ../dev_handler/message.o ../runtime_util/runtime_util.o ../logger/logger.o
+VIRTUAL_DEVICES := $(shell ls client/virtual_devices/*Device.c | awk -F'.' '{ print $$1 }')
```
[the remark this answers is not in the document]
AVSurfer123: Uhh idk. I think its something to do with combining shell + Makefile variables. I just used this somewhere else and it worked so I copy pasted it here. It will output everything before the first `.`, so `client/virtual_devices/GeneralTestDevice` which is used to list all devices to make.

runtime
```diff
@@ -16,10 +16,11 @@ cd $RUNTIME_DIR
...
     build)
-        pushd shm_wrapper && make && popd
-        pushd net_handler && make && popd
-        pushd executor && make && popd
-        pushd dev_handler && make && popd
+        shift 1
+        pushd shm_wrapper && make $@ && popd
+        pushd net_handler && make $@ && popd
+        pushd executor    && make $@ && popd
+        pushd dev_handler && make $@ && popd
```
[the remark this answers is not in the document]
AVSurfer123: oh this was mainly just so I can do `runtime build -B` to force a rebuild. I didn't think about using it to build only 1 sub target. But I got what you said to work by removing `set -e` and changing `popd`.

net_handler/udp_conn.c
levincent06: I didn't comment on this enum thing because I wasn't sure what the best solution was, but what Daniel and Ben mentioned above sounds good to me!

runtime
```diff
@@ -39,9 +40,9 @@ case $1 in
     clean)
```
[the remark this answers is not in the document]
AVSurfer123: yea. or just delete the git repo and re clone it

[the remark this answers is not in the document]
AVSurfer123: `clean` will only remove files that `build` or `run` will make

scripts/test.sh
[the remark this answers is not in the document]
AVSurfer123: Yea what you guys said makes sense for the Travis CI. I was mostly just talking about testing 1 specific file. But its probably not much of an issue so I'll just leave it.
@benliao1  your suggestion would make it recompile the test every time you wanted to run it, which I'm not sure if we want.

executor/runtime.pxd
```diff
@@ -4,6 +4,7 @@ cdef extern from "../runtime_util/runtime_util.h":
     int MAX_DEVICES
     int MAX_PARAMS
     int NUM_GAMEPAD_BUTTONS
+    int NUM_KEYBOARD_BUTTONS
```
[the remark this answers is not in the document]
AVSurfer123: Yes it needs to be updated. Its pretty simple and you can't really forget cus otherwise you can't use runtime in Cython lol. I don't think there's anyway to automate this.

shm_wrapper/shm_start.c
```diff
@@ -127,14 +127,17 @@ int main() {
         dev_shm_ptr->net_sub_map[i] = -1;
         dev_shm_ptr->exec_sub_map[i] = 0;
     }
-    gp_shm_ptr->buttons = 0;
-    for (int i = 0; i < 4; i++) {
-        gp_shm_ptr->joysticks[i] = 0.0;
+    for (int j = 0; j < 2; j++) {
```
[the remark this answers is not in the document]
AVSurfer123: actually thinking about it, having more than 2 sources would break a lot of the code. so I don't see the need to have a define statement. similar to only having data and command streams for devices

shm_wrapper/shm_wrapper.c
AVSurfer123: well its gonna get deleted in your PR so its fine lol

shm_wrapper/shm_wrapper.c
AVSurfer123: yes

net_handler/udp_conn.c
AVSurfer123: Hmm ok I wanted to make it general but a ternary condition would be cleaner if we are sticking with only 2 sources.

The main problem is `input->source` is a Protobuf enum corresponding to 0 and 1. So I need to convert the protobuf enum to a Runtime robot_desc enum. I do this in several places and so didn't want to use a if/switch case.

shm_wrapper/shm_wrapper.c
AVSurfer123: This is for a different function, the `Robot.log` to log custom data to Dawn. They `key` here is something the students choose to describe their data. I just added this cus I realized my previous implementation didn't correctly check for an error. Its not related to the main PR on inputs at all.

tests/client/net_handler_client.c
AVSurfer123: yes, good catch

AVSurfer123: actually, nvm it doesn't matter since `exit` will cause the OS to automatically free all memory for the process

executor/studentapi.pyx
```diff
@@ -107,6 +107,39 @@ cdef class Gamepad:
         raise KeyError(f"Invalid gamepad parameter {param_name}")
 
 
+cdef class Keyboard:
+    """
+    The API for accessing the keyboard.
+    """
+    cdef bint available
```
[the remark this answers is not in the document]
AVSurfer123: its a boolean int. its needed to treat C ints as Python booleans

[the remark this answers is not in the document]
AVSurfer123: I think there's some more to it but you'd need to read the Cython docs for that

executor/studentapi.pyx
```diff
@@ -107,6 +107,39 @@ cdef class Gamepad:
...
+
+    def __cinit__(self):
+        """Initializes the mode of the robot. """
+        self.available = robot_desc_read(RUN_MODE) == TELEOP
+
+    cpdef get_value(self, str param_name):
```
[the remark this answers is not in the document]
AVSurfer123: Maybe? I just kept it as `get_value` to be consistent with the other 2 API classes. But we can change them if you want. If it were me, I would rewrite the API to be more Pythonic and not use strings like we have, and instead have functions or attributes that exactly give what you want. So like `Gamepad.get_joystick_x()` or `Gamepad.BUTTON_A`.

executor/studentapi.pyx
```diff
@@ -193,6 +226,8 @@ cdef class Robot:
             value: value of the parameter. Must be an int, float, or bool.
         
         """
+        if len(key) >= 64:
+            raise ValueError(f"Cannot log parameter {key} since it is more than 63 characters long")
```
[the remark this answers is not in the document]
AVSurfer123: Yea this is just another error check I forgot about. Added it to runtime_util

runtime
```diff
@@ -56,8 +57,12 @@ case $1 in
         scripts/update.sh
         ;;
 
+    stop)
+        scripts/stop.sh
```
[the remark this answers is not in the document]
AVSurfer123: Its helpful in 2 scenarios. 

1. The one I was facing was something would segfault and none of the process would shut down from the testing clients. So I created `runtime stop` to force kill them for me.
2. Ctrl+C in each terminal takes too long lol

shm_wrapper/Makefile
```diff
@@ -18,4 +18,4 @@ shm_stop: shm_stop.c $(FILES)
 	gcc $^ -o $@ $(LIBS)
 
 clean:
-	rm test1 test2 shm_start shm_stop
+	rm -f test1 test2 shm_start shm_stop
```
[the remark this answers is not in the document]
AVSurfer123: the `-f` is everywhere if you see the other Makefiles. It is necessary to have `make` not give us errors when the file doesn't exist.

shm_wrapper/shm_wrapper.c
```diff
@@ -784,26 +799,32 @@ int gamepad_read(uint32_t* pressed_buttons, float joystick_vals[4]) {
     my_sem_post(rd_sem, "robot_desc_mutex");
 
     // wait on gp_sem
-    my_sem_wait(gp_sem, "gamepad_mutex");
+    my_sem_wait(input_sem, "input_mutex");
 
-    *pressed_buttons = gp_shm_ptr->buttons;
+    int index = source - GAMEPAD;  // Assumes KEYBOARD is after GAMEPAD in robot_desc_field_t enum
```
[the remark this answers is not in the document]
AVSurfer123: yea the ternary condition is cleaner if we stick to only having 2 input sources

shm_wrapper/shm_wrapper.c
AVSurfer123: Theres no need for a check as we don't read the higher bits anyways in the student API. Whoever is sending the data is responsible for checking the bitmap in your example, not Runtime.

tests/client/executor_client.h
AVSurfer123: Imo you shouldn't include `runtime_util` just for headers as it masks what is really needed in the source file. One thing I hate about C/C++ is you don't know where a function/type is defined just by looking at the source file. You should only include `runtime_util` if you need some of the functions/types specific to Runtime defined in it.

tests/test.c
AVSurfer123: yes, its only used to print buttons.

tests/logger.config
AVSurfer123: This is the test logger config. I made it DEBUG so I could actually see my debug statements print during the test. I think this should be the default for testing.

runtime_util/runtime_util.h
AVSurfer123: hmmmm sounds really bad. It's really ugly to use an enum for that. I'd rather use a function (same for gamepad) that uses our predefined arrays.

AVSurfer123: ok i added some functions that are much nicer

executor/studentapi.pyx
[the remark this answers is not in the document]
benliao1: Ok sounds good,  yeah I read up on some Cython docs after I made the comment and it makes more sense to me~~

executor/studentapi.pyx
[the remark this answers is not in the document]
benliao1: Yea I consulted Charles over at Pimulator today and I think there's consensus over there about keeping it `get_value`, for consistency like you said! So let's keep it `get_value()` for now~~

runtime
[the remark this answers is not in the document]
benliao1: Mm okay that's fine I guess, the second point though 😂 
As long as it works and doesn't unintentionally kill any random processes then we should be fine.

tests/client/executor_client.h
benliao1: Ok that's a fair point, just the opposite of what I've been doing since high school though so if you want to change that we should keep it consistent and do it for all of Runtime :\\
Are you saying that we should not include any headers in `runtime_util` (or only the standard library headers that are actually used in `runtime_util`) and for each file include things that are necessary in that file only?

tests/logger.config
benliao1: Mm if we still do print output comparisons in tests then I hesitate to change this, @levincent06 do you know if any of our tests still use output comparisons?

shm_wrapper/shm_wrapper.h
```diff
@@ -310,7 +325,8 @@ int gamepad_write(uint32_t pressed_buttons, float joystick_vals[4]);
  *    value: value of the parameter, with the corresponding type filled with data
  * Returns:
  *    0 on success
- *    -1 on failure (maximum number of custom parameters is reached so this parameter isn't written)
+ *   -1 when maximum number of custom parameters is reached
+ *   -2 when the given key is longer than 63 characters
```
levincent06, line 329: Can we expect that 63 will be "hardcoded"? I can see that 63 is pretty sizable so it probably won't get any larger.

tests/integration/tc_150_1.c
```diff
@@ -0,0 +1,40 @@
...
+    send_run_mode(DAWN, TELEOP);
+    sleep(.1);
+
+    buttons = 1 << 24;  // 'y'
+    send_user_input(buttons, garbage, KEYBOARD);
+    sleep(.1);
+    check_inputs(buttons, garbage, KEYBOARD);
+    sleep(1);
+    add_ordered_string_output("1");
+    add_ordered_string_output("1");
```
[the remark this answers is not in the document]
levincent06: Why do we need to check twice?

runtime_util/runtime_util.h
```diff
@@ -24,9 +24,12 @@
 
 #define DEVICES_LENGTH 64  // The largest device type number + 1.
 
-#define NUM_DESC_FIELDS 5  // Number of fields in the robot description
+#define NUM_DESC_FIELDS 6  // Number of fields in the robot description
 
-#define NUM_GAMEPAD_BUTTONS 17  // Number of gamepad buttons
+#define LOG_KEY_LENGTH 64  // Max length of a key for custom Robot.log data
```
levincent06, line 29: Can we name this `MAX_LOG_KEY_LENGTH`

runtime
```diff
@@ -8,18 +8,21 @@
...
 #     runtime test integration/tc_68_2   (to run the automated test tc_68_2)
 
-set -e  # Makes any errors exit the script
-
 RUNTIME_DIR=$(dirname "${BASH_SOURCE[0]}")
 
 cd $RUNTIME_DIR
 
 case $1 in
     build)
```
levincent06, line 16: I remember making a comment about this, but it seems to have gotten resolved.. what was the rationale for having arguments to `runtime build`?

shm_wrapper/shm_wrapper.h
levincent06: JK this looks like this is `MAX_LOG_KEY_LENGTH - 1`? If so then we should reference that `#define` instead of mentioning 63, which may change

levincent06 (changes requested):

tests/logger.config
levincent06: There are still tests that use output comparisons, specifically those where we test for output that doesn't get stored in shared memory. For example, see `tc_71_2.c` where we test that the student code properly has a traceback error (one divided by zero) We also have @0neDividedbyZer0's  (how fitting lol) regex addition for future output comparisons, if needed.

There's also `tc_71_1` which tests for start position through output, but that was an oversight and I'll add direct shared memory checking in another PR, specifically addressing issue #155 

If I'm understanding correctly, the issue with lowering the log level to `DEBUG` is that the debugs may give us a false positive for running tests. IMO this is an issue with output checking in general, which the newer test API tried to address. :\

I think the options are
1) Keep it as `INFO` logs, as they should more accurately reflect the state of Runtime, or
2) Test cases that do output checks should be very specific about what outputs are expected, though this may require using precise regex

tests/integration/tc_150_1.c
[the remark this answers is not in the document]
AVSurfer123: cus I print it at least twice in the student code. makes sure the sleep timing is working correctly

runtime
```diff
@@ -56,8 +57,12 @@ case $1 in
         scripts/update.sh
         ;;
 
+    stop)
+        scripts/stop.sh
+        ;;
+
     *)
-        commands="build run test format clean flash update"
+        commands="build run test format clean flash update stop"
```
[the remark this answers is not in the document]
AVSurfer123: That should probably be a separate PR. This one is already bloated

scripts/stop.sh
```diff
@@ -0,0 +1,11 @@
...
+printf "first here\n"
+pkill executor
+printf "second\n"
+pkill dev_handler
+printf "third\n"
+pkill net_handler
+printf "fourth\n"
+cd shm_wrapper && ./shm_stop
```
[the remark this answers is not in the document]
AVSurfer123: Discussed above

tests/client/executor_client.h
AVSurfer123: uhh, I don't care too much about that, though that would probably be simpler. I just think you shouldn't import a file just for its headers. In the rest of the project, we only import `runtime_util` if we need its attributes

tests/logger.config
AVSurfer123: how could `DEBUG` statements ever cause false positives? if they do then the output check is way too broad

shm_wrapper/shm_wrapper.h
AVSurfer123: yea I just forgot to update the doc

runtime
AVSurfer123: I answered on the previous comment about it. one of the beginning ones

runtime_util/runtime_util.h
AVSurfer123: that gets pretty long, i think its fine as is

tests/logger.config
benliao1: One possible way that we could get false positives is if we're checking for something that should appear 5 times but there are DEBUG statements that make it so the output appears more than that? 
I agree that the test cases should ideally not rely on output comparison but I don't really see a way of doing all of that. The test framework overhaul we did at the beginning of the semester went a long way to address that though.

shm_wrapper/shm_wrapper.c
```diff
@@ -837,12 +859,17 @@ int log_data_write(char* key, param_type_t type, param_val_t value) {
...
 
-    // return if we didn't find the log data corresponding to this key
+    // return if we ran out of keys for log data
     if (idx == UCHAR_MAX) {
         log_printf(ERROR, "Maximum number of %d log data keys reached. can't add key %s", UCHAR_MAX, key);
         return -1;
     }
 
+    if (strlen(key) >= LOG_KEY_LENGTH) {
+        log_printf(ERROR, "Key name %s for log data is longer than 64 characters", key);
```
levincent06, line 869: A string of length `LEN` has `LEN + 1` characters (null terminator), so is `LOG_KEY_LENGTH` the upper bound for the key string? Also we can use the `#define` in the log as well.

shm_wrapper/shm_wrapper.h
```diff
@@ -310,7 +325,8 @@ int gamepad_write(uint32_t pressed_buttons, float joystick_vals[4]);
  *    value: value of the parameter, with the corresponding type filled with data
  * Returns:
  *    0 on success
- *    -1 on failure (maximum number of custom parameters is reached so this parameter isn't written)
+ *   -1 when maximum number of custom parameters is reached
+ *   -2 when the given key is longer than LOG_KEY_LENGTH characters
```
levincent06, line 329: See my comment in `shm_wrapper.c`, but imo it might be clearer if it's phrased as
`...when the length of the key is longer than LOG_KEY_LENGTH` 
or if the null-terminator is involved, 
`...when the length of the key is longer than LOG_KEY_LENGTH - 1` 
C makes character count a bit annoying

tests/integration/tc_150_1.c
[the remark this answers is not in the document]
levincent06: I see. can you add a comment mentioning that it prints out `1` in the main loop, which loops twice? I was a bit confused

levincent06 (changes requested):

shm_wrapper/shm_wrapper.c
AVSurfer123: its the max length including the null byte.

tests/logger.config
AVSurfer123: idk your example doesn't make sense. why would the debug statements have the output that you are checking for? if that occurs, you can always remove the debug statement since you added it in the first place.

shm_wrapper/shm_wrapper.h
AVSurfer123: it should be the first one when you include the null byte in the count.

shm_wrapper/shm_wrapper.h
```diff
@@ -322,6 +338,6 @@ int log_data_write(char* key, param_type_t type, param_val_t value);
  *    types: array that will be filled with the parameter types, up to `num_params`
  *    values: array that will be filled with the parameter values, up to `num_params`
  */
-void log_data_read(uint8_t* num_params, char names[UCHAR_MAX][64], param_type_t types[UCHAR_MAX], param_val_t values[UCHAR_MAX]);
+void log_data_read(uint8_t* num_params, char names[UCHAR_MAX][LOG_KEY_LENGTH], param_type_t types[UCHAR_MAX], param_val_t values[UCHAR_MAX]);
```
levincent06, line 341: Last change! The comment for the argument `names` should be updated, similar to the one in `log_data_write()`. Something along the lines of `at most LOG_KEY_LENGTH characters long, including null-terminator`?

levincent06 (changes requested):

tests/logger.config
DanielMolina24: We have two functions that do string checks that we currently use for checking different info such as headers for devices or that the "right" error is outputted . Theres one for strings in a certain order vs one for just checking all the strings in general unordered. I believe the example Ben is giving is if we trigger an error two times in a row and then check that an error occurs twice in a row using the add_ordered_string function, we might get a false positive  from the DEBUG statement right after the first error pops up. It'll see that the error string is outputted two times in a row and think the test was successful, but it wouldn't get around to checking the output of the second triggered error since the first error appeared two times, giving us that false positive

shm_wrapper/shm_wrapper.h
AVSurfer123: It doesn't need it. There's no pointers here, just statically allocated arrays. this function will never fail unlike the write function

tests/logger.config
AVSurfer123: but why would the debug statement be printing out the same error as another log statement? the debug logs are things you add for testing so you wouldn't choose it to be the same statement. these examples seem very contrived and wouldn't really ever happen

shm_wrapper/shm_wrapper.h
levincent06: Even though it will never fail, the comment is outdated, right? It still says `63` and it's better to refer to the `#define` that it's actually referring to, in the case that it will be changed. (I meant to comment this for line 337, but GitHub doesn't seem to allow me to add comments to unchanged lines.)

AVSurfer123: Ok yea I was confused. I'll fix it

benliao1 (approved):
Looks good! We should work on updating the Wiki page too as soon as possible, but that is after this is merged~~

levincent06 (approved):

DanielMolina24 (approved):

PR #160, opened 2021-01-11
[TEST] Fixed latency bug

### LINKED ISSUE #156: [MISC] Latency test tc_71_10 has invalid results
In `tc_71_10`, we press a button which writes to the TImeTestDevice in the student code. When the virtual device receives this write, it will write a timestamp to one of its parameters. This timestamp can be read off of shared memory to calculate the latency.

See the following files:
- Student code: `tests/student_code/runtime_latency.py`
- TimeTestDevice: `tests/client/virtual_devices/time_dev.c`

The test case itself will always pass. There is no check for what the latency should be. In the past, the output is usually 0 or 1 milliseconds. The error is that the parameter `TIMESTAMP` is read as 0 in the test case, so the latency will appear to be negative. This error has been observed only recently, and it's not clear exactly what change may have caused this to happen.

We should
1. Investigate why `TIMESTAMP` is 0
2. Add a check in `tc_71_10` to pass if and only if the latency is in [0, 5)

DanielMolina24:
Resolves #156. It seems that the bug can be tracked down to commit #8213292. A removal of a `sleep(40000)` inside the `send_gamepad_state` function of `net_handler_cli` was what caused the issue. Inside the test case, we send a gamepad state where button A is pressed and immediately follow with sending another gamepad state with button A unsent. I believe what may have happened was that perhaps it probably wasn't done processing the first gamepad_state received allowing for the virtual device to update its timestamp before the second gamepad state was received, or it received them out of order, which could explain why the test worked successfully like 1/20 times that I ran the test yesterday. I'm honestly not too sure if a bug like this can happen with real robots and gamepads/keyboards since this seemed more like a client issue. Either way this was an easy fix
- Added sleep after sending updated gamepad state in `tc_71_10.c`
- Added a function to test framework to allow us to check the latency of a provided start time and a parameter of a device


tests/test.c
```diff
@@ -647,3 +647,28 @@ void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_typ
     }
     print_pass();
 }
+
+// Returns if the latency is neither too high nor too low
+void check_latency(char* dev_name, uint64_t uid, char* param_name, int32_t lower_bound_latency, int32_t upper_bound_latency, int32_t start_time) {
```
levincent06, line 652: I think it's probably fine to hard code the "lower bound" as 0 here. We shouldn't punish our system for being too fast! Though checking that it's at least 0 is a validity check because negative latency means something is wrong.

tests/test.c
```diff
@@ -647,3 +647,28 @@ void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_typ
...
+    uint32_t param_idx = (uint32_t) get_param_idx(dev_type, param_name);
+    device_read_uid(uid, EXECUTOR, DATA, (1 << param_idx), vals_after);
+    param_val_t received = vals_after[param_idx];
+    int32_t elapsed_time = received.p_i - start_time;
+    fprintf(stderr, "Time received is %d\n", received.p_i);
+    if(elapsed_time >= upper_bound_latency && elapsed_time < lower_bound_latency){
+        print_fail();
+        fprintf_delimiter(stderr, "Expected:");
+        fprintf(stderr, "%s >= %d\n", param_name, lower_bound_latency);
+        fprintf(stderr, "%s < %d\n", param_name, upper_bound_latency);
```
levincent06, line 666: Minor nitpick but let's make this one line so it's easier to read
`fprintf(stderr, "0 <= %s < %d\n", param_name, upper_bound_latency)`
(Assuming we're hard-coding the 0)

tests/test.c
```diff
@@ -647,3 +647,28 @@ void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_typ
...
+// Returns if the latency is neither too high nor too low
+void check_latency(char* dev_name, uint64_t uid, char* param_name, int32_t lower_bound_latency, int32_t upper_bound_latency, int32_t start_time) {
+    uint8_t dev_type = device_name_to_type(dev_name);
+    device_t* dev = get_device(dev_type);
+
+    param_val_t vals_after[dev->num_params];
+    uint32_t param_idx = (uint32_t) get_param_idx(dev_type, param_name);
+    device_read_uid(uid, EXECUTOR, DATA, (1 << param_idx), vals_after);
+    param_val_t received = vals_after[param_idx];
+    int32_t elapsed_time = received.p_i - start_time;
```
levincent06, line 660: I think `received` can be renamed to `end_time` for clarity

tests/test.h
```diff
@@ -149,4 +149,15 @@ void same_param_value(char* dev_name, uint64_t uid, char* param_name, param_type
...
+*Checks the latency between an action and the parameter it affects being changed
+ * Arguments:
+ *    dev_name: The name of the device whose parameters are being checked
+ *    uid: uniquie identifier of device
+ *    param_name: The name of the parameter being compared
+ *    lower_bound_latency: The expected lowerbound latency 
+ *    upper_bound_latency: The expected upperbound latency
+ *    start_time: The time right before the button is pressed
+*/
+void check_latency(char* dev_name, uint64_t uid, char* param_name, int32_t lower_bound_latency, int32_t upper_bound_latency, int32_t start_time);
```
levincent06, line 162: I think adding this function is a good idea! From what I can tell though, this function will be used only for devices with some kind of `TIMESTAMP` parameter like the `TimeTestDevice` has, right?

We should add a comment explaining that the parameter has to be specifically something that holds a timestamp to be used as the "end time"

DanielMolina24: Good idea, I'll add in a comment on that

tests/test.c
```diff
@@ -647,3 +647,27 @@ void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_typ
     }
     print_pass();
 }
+
+// Returns if the latency is neither too high nor negative
+void check_latency(char* dev_name, uint64_t uid, char* param_name, int32_t upper_bound_latency, int32_t start_time) {
+    uint8_t dev_type = device_name_to_type(dev_name);
+    device_t* dev = get_device(dev_type);
```
AVSurfer123, line 654: Need to check for errors as a result of both of these functions.

tests/test.c
```diff
@@ -647,3 +647,27 @@ void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_typ
...
+
+    param_val_t vals_after[dev->num_params];
+    uint32_t param_idx = (uint32_t) get_param_idx(dev_type, param_name);
+    device_read_uid(uid, EXECUTOR, DATA, (1 << param_idx), vals_after);
+    param_val_t end_time = vals_after[param_idx];
+    int32_t elapsed_time = end_time.p_i - start_time;
```
AVSurfer123, line 660: This function only seems useful if its a `TimeTestDevice`. Is there any point to requiring a `dev_name` and `param_name` then?

AVSurfer123 (changes requested):

tests/test.c
DanielMolina24: Oh right! Good catch!

tests/test.c
DanielMolina24: This was there just in case in the future someone decides to write a new virtual device with a `TIMESTAMP` on it. I do agree though, the `param_name` is kind of redundant

levincent06: Yea this is interesting and didn't comment on this because we may want future devices with `TIMESTAMP` params. Though I'm having trouble thinking of a scenario when we would need a new device.

DanielMolina24: Yea thats a good point, they can easily just connect the `TimeTestDevice`. Its an easy fix to the function if they do write a new device in the future. I'll remove it from there

tests/test.h
```diff
@@ -149,4 +149,12 @@ void same_param_value(char* dev_name, uint64_t uid, char* param_name, param_type
  */
 void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_type_t param_type, param_val_t expected_low, param_val_t expected_high);
 
+/**
+* Checks the latency between an action and the TIMESTAMP parameter on a TimeTestDevice
+ * Arguments:
+ *    uid: uniquie identifier of device
+ *    upper_bound_latency: The expected upperbound latency
+ *    start_time: The time right before the button is pressed
```
levincent06, line 157: Should specify that the `uid` is specifically for a `TimeTestDevice` (ik it's repetitive but it makes it clearer). Also typo :P 

`start_time` Mentions a button, though in the future it may not be a button press, so the function in general calculates the time difference between a `TimeTestDevice`'s `TIMESTAMP` and the provided "start" timestamp, which can be anything!

levincent06 (changes requested):

tests/integration/tc_71_10.c
```diff
@@ -28,9 +29,11 @@ int main() {
     send_run_mode(SHEPHERD, TELEOP);
 
     // Start the timer and press A
-    int32_t start = millis() % 1000000000;  // 9 digits, just like TimeTestDevice
-    buttons |= get_button_bit("button_a");
+
+    uint64_t start = millis();
+    buttons |= (1 << BUTTON_A);
```
levincent06, line 34: `buttons |= get_button_bit("button_a");` should be ok. Don't break abstraction barrier

tests/test.c
```diff
@@ -649,3 +663,46 @@ void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_typ
...
+    if(success < 0){
+        fprintf(stderr, "Error reading device with UID: %llu", uid);
+    }
+    
+    // shorten to last 9 digits, like the TimeTestDevice's TIMESTAMP
+    int32_t start_time_shortened = start_time  % 1000000000; 
+
+    param_val_t end_time = vals_after[param_idx];
+    int32_t elapsed_time = end_time.p_i - start_time_shortened;
+    fprintf(stderr, "Time received is %d\n", end_time.p_i);
```
levincent06, line 696: Maybe we can print out `elapsed_time` instead, which will be more useful

tests/test.c
```diff
@@ -649,3 +663,46 @@ void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_typ
...
+    if(elapsed_time >= upper_bound_latency && elapsed_time < 0){
```
levincent06, line 697: I think you want an `||` instead of `&&`

tests/test.h
```diff
@@ -150,4 +150,12 @@ void same_param_value(char* dev_name, uint64_t uid, char* param_name, param_type
...
 void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_type_t param_type, param_val_t expected_low, param_val_t expected_high);
 
+/**
+* Checks the latency between an action and the TIMESTAMP parameter on a TimeTestDevice
+ * Arguments:
+ *    uid: unique identifier of TimeTestDevice
+ *    upper_bound_latency: The expected upperbound latency
+ *    start_time: The start of a timer provided by the test case(in ms)
+*/
+void check_latency(uint64_t uid, int32_t upper_bound_latency, uint32_t start_time);
```
levincent06, line 160: Hm so in the test case, you pass in a call to `millis()` as the start time, where `millis()` returns a `uint64_t`.

You can make the parameter type `uint64_t` as well, and mention in the function comment that `start_time` should be some call to `millis()`

levincent06 (changes requested):

tests/integration/tc_71_10.c
DanielMolina24: Oh whoops yea when resolving a merge conflict I probably accepted the wrong one. Let me fix that

tests/test.c
DanielMolina24: Yea missed that

tests/test.h
DanielMolina24: Oh yea I forgot to have the `.h` reflect that change. Thanks for the catch

levincent06 (approved):

tests/integration/tc_71_10.c
```diff
@@ -28,25 +29,19 @@ int main() {
     send_run_mode(SHEPHERD, TELEOP);
 
     // Start the timer and press A
-    int32_t start = millis() % 1000000000;  // 9 digits, just like TimeTestDevice
+    uint64_t start = millis();  // 9 digits, just like TimeTestDevice
```
AVSurfer123, line 32: Fix comment

tests/test.c
```diff
@@ -603,11 +603,25 @@ void same_param_value(char* dev_name, uint64_t uid, char* param_name, param_type
...
     uint8_t dev_type = device_name_to_type(dev_name);
+    if(dev_type == -1){
+        print_fail();
+        fprintf(stderr, "%s type not found\n", dev_name);
+        end_test();
+        exit(1);
+    }
     device_t* dev = get_device(dev_type);
-
+    if(dev == NULL){
```
AVSurfer123, line 613: this should never fail since the previous `dev_type` isn't -1

AVSurfer123 (changes requested):

tests/test.c
AVSurfer123: You removed it in `check_latency`, but it also can be removed here in `check_param_range`.

tests/test.c
```diff
@@ -649,3 +663,52 @@ void check_param_range(char* dev_name, uint64_t uid, char* param_name, param_typ
...
+        fprintf(stderr, "Error reading device with UID: %llu", uid);
+    }
+    
+    // shorten to last 9 digits, like the TimeTestDevice's TIMESTAMP
+    uint32_t start_time_shortened = start_time  % 1000000000; 
+    param_val_t end_time = vals_after[param_idx];
+    
+    int32_t elapsed_time = end_time.p_i - start_time_shortened;
+    fprintf(stderr, "Elapsed time is %d\n", end_time.p_i);
+    if(elapsed_time >= upper_bound_latency || elapsed_time < 0){
```
AVSurfer123, line 697: if `elapsed_time < 0`, doesn't it mean there is some other error going on? we should handle it differently to print out a different error msg.

DanielMolina24: Hmm yea I see your point. Let me fix that real quick

AVSurfer123 (approved):

PR #166, opened 2021-01-15
 [TEST]Merged net_handler_cli and keyboard_interface into a single process

### LINKED ISSUE #129: [CLI] Merge keyboard interface and net handler cli
Currently, keyboard interface acts as a net handler cli stuck in TELEOP mode. (You can't switch to AUTO mode in keyboard interface).  It would be nice to have it so that when we switch run mode to TELEOP mode in net handler cli, it "becomes" the keyboard interface, accepting keyboard inputs from `keyboard_inputs.py`. There should also be the option to switch to another run mode from TELEOP.

Requires changes in:
- keyboard_interface.c
- net_handler_cli
- (Possibly) keyboard_inputs.py

DanielMolina24:
- `net_handler_cli` spawns a thread that runs the `keyboard_interface` main function in a loop 
 - Removed "prompt_gamepad_state()" from `net_handler_cli` as the keyboard can now easily be used during teleop 
 - Edited makefile to no longer produce an executable for `keyboard_interface`
 - Added an extra bitstring for keyboard inputs to send that composes of the keys pressed on the keyboard 
 - `keyboard_interface` takes into account the new bitstring and parses it before passing it into `send_user_input()` function
 Resolves #129


tests/cli/keyboard_interface.c
```diff
@@ -59,63 +53,61 @@ int connect_tcp() {
...
 
     if ((bind(sockfd, (struct sockaddr*) &servaddr, sizeof(servaddr))) != 0) {
-        printf("socket bind failed... %s\n", strerror(errno));
+        printf("Keyboard interface: socket bind failed... %s\n", strerror(errno));
         exit(0);
     }
 
-    printf("Waiting for client\n");
+    printf("Keyboard interface: Keyboard waiting for client\n");
+    sleep(2);
```
levincent06, line 61: Is the `sleep(2)` necessary?

tests/cli/keyboard_interface.c
```diff
@@ -50,7 +44,7 @@ int connect_tcp() {
 
     sockfd = socket(AF_INET, SOCK_STREAM, 0);
     if (sockfd == -1) {
-        printf("socket creation failed...\n");
+        printf("Keyboard interface: socket creation failed...\n");
```
levincent06, line 47: Just a convention thing--error messages should follow the format
`"<function>: <error message>"`
This is so we know exactly in which function caused the error.
So maybe `"connect_tcp: socket creation failed in Keyboard interface"` or similar?

Just realized we didn't add this to the Wiki yet hm

tests/cli/keyboard_interface.c
```diff
@@ -59,63 +53,61 @@ int connect_tcp() {
...
-
-    start_net_handler();
-    sleep(2);
-    // Connect dummy gamepad and start teleop mode
-    send_user_input(0, joystick_vals, GAMEPAD);
-    send_run_mode(DAWN, TELEOP);
-    sleep(2);
+    uint64_t keyboard_buttons = 0;
+    // The received keyboard bitstring from Python
+    char keyboard_buff[48];
```
levincent06, line 87: In `keyboard_inputs.py`, is the gamepad bitmap 32 bits or `NUM_GAMEPAD_BUTTONS` bits when sent over the socket?

It seems to me that `char gamepad_buff[33]` is based off of 32-bits, while `char keyboard_buff[48]` is based off of `NUM_KEYBOARD_BUTTONS`. We should keep them consistent!

tests/cli/keyboard_interface.c
```diff
@@ -59,63 +53,61 @@ int connect_tcp() {
...
-
     // Receive a bitstring of buttons and parse it
     while (1) {
         // Reset buttons/joysticks
-        buttons = 0;
-        memset(buff, 0, 32 * sizeof(char));
+        gamepad_buttons = 0;
+        memset(gamepad_buff, 0, 32 * sizeof(char));
         memset(joystick_vals, 0, 4 * sizeof(float));
         // Read in bit string of size 32, not sizeof(buf)(33) since we are being sent a 32 len string
```
levincent06, line 98: See my comment above about the size of the bitmap being sent

tests/cli/keyboard_interface.c
```diff
@@ -148,15 +140,21 @@ int main() {
...
-                buttons |= (1 << i);
+            if (gamepad_buff[i] == '1') {
+                gamepad_buttons |= (1 << i);
+            }
+        }
+
+        // Set bitmap for keyboard
+        for(int i = 0; i < NUM_KEYBOARD_BUTTONS; i++) {
+            if(keyboard_buff[i] == '1') {
+                keyboard_buttons |= (1 << i);
```
levincent06, line 153: Can't remember if I said it, but I take back what I said a long time ago about the for loop :P
It makes sense to do
```
for (int i = 0; i < NUM_GAMEPAD_BUTTONS; i++)
```
like you did for the keyboard buttons, in case the order changes for whatever reason.

tests/client/net_handler_client.c
```diff
@@ -307,6 +307,12 @@ void start_net_handler() {
             printf("pthread_create: output dump\n");
         }
         usleep(400000);  // allow time for thread to dump output before returning to client
+        
+        // execute the keyboard_interface on a seperate thread
+        pthread_t keyboard_id; // id of thread running the keyboard_interface
+        if(pthread_create(&keyboard_id, NULL, (void *)setup_keyboard, NULL) != 0) {
+            printf("pthread create: setup keyboard");
```
benliao1, line 314: Did you test that this works? Maybe I'm dumb but I can't seem to find the function `setup_keyboard` that is supposed to be running in this thread

benliao1 (commented):
The changes make sense (I didn't test it myself) except the one comment I have--where is `setup_keyboard` function?

tests/client/net_handler_client.c
```diff
@@ -307,6 +307,12 @@ void start_net_handler() {
...
+        }
```
levincent06, line 315: Maybe put this in another function, because almost all the other logic in this function deals with setting up net handler itself

tests/client/net_handler_client.h
```diff
@@ -87,4 +87,10 @@ void print_next_dev_data();
  */
 void update_tcp_output_fp(char* output_address);
 
+/**
+* Setups connection to keyboard_inputs python script. Processes incoming inputs and translates
+* it to gamepad and joystick info to send through the send_gamepad_state() function
+*/
+void setup_keyboard();
```
levincent06, line 94: I think you meant to add this to a new file `keyboard_interface.h`? I'm a little confused about the file dependencies actually.

- `net_handler_client` -- Functions to start/stop net handler, and send inputs to net handler; Depends on `net_util.h`
- `net_handler_cli` -- Prompts user input; Depends on `net_handler_client`
- `keyboard_interface` -- Opens a socket to accept connection from `keyboard_inputs.py`; Depends on `net_handler_client`

Maybe `keyboard_interface` can be put completely in `net_handler_client`, so we follow the same structure that all the CLIs gives prompts to call client functions?

levincent06 (changes requested):

tests/client/net_handler_client.c
levincent06: The function is implemented in `keyboard_interface.c`, but declared in `net_handler_client.h`. I think it's just a mistake :P I made a comment about this in my review below: maybe the "keyboard interface" functions can be merged with net handler client?

tests/cli/keyboard_interface.c
DanielMolina24: This was just so that the print statements from the keyboard interface dont mix around with the cli's print statements which is what it was doing w/o the sleep

tests/cli/keyboard_interface.c
DanielMolina24: Can you explain what you mean? They are 33 and 48 to keep account of the NULL terminator at the end, if you mean that I should use the defined `NUM_GAMEPAD_BUTTONS` + 1 and `NUM_KEYBOARD_BUTTONS`+1 then yea I can do that

tests/client/net_handler_client.c
DanielMolina24: Yea so I did test it and it was on purpose. So I had thought about moving the functions over to net handler client but kind of wanted to leave it in its own file since it does its own communicating with the python script and processing that the `net_handler_client` doesn't do. I also wasn't the biggest fan of making a `.h` for like one function that would be in it. So in the Makefile I added the keyboard_interface under the net_handler_cli's rules:
`net_handler_cli: cli/net_handler_cli.* client/net_handler_client.* cli/keyboard_interface.* ../net_handler/net_util.* \`

tests/client/net_handler_client.h
DanielMolina24: See comment above

tests/cli/keyboard_interface.c
levincent06: Ok I see. Yeah we should use `#define`'s where possible. I was confused because `NUM_GAMEPAD_BUTTONS` is 17. With the joysticks as possible "buttons", there still aren't 32 buttons from my understanding. I wasn't sure where `32 + 1` came from

tests/cli/keyboard_interface.c
levincent06: Can you move the `sleep()` to the CLI instead? Since this sleep is more of an "aesthetic" thing, it's more appropriate for the CLI

tests/client/net_handler_client.h
DanielMolina24: Hmm actually I might just end up putting it in a .h file to just clean up the stuff in there such as enums and includes

tests/client/net_handler_client.c
DanielMolina24: I might just end up putting it in its own `.h` right now though to clean up the `keyboard_interface` file since it also has enums and some include statements in there

levincent06: I haven't seen functions declared like this before--I would have made a header file for just that one function. Maybe Ben can comment on this because he has that K&R experience

benliao1: I think with this change, the intention was for the keyboard interface to be merged with the CLI right?
You're right, in my experience, a function should always be declared in a header file with the same way as the source file in which it is defined. So we should definitely change this as the way it is.
Functionally, I think the keyboard interface stuff is more of a CLI thing (since it's used for parsing input from the command-line, kinda) which it then passes on to the `net_handler` process through `net_handler_client`. So in my view, you're right, the `net_handler_client` probably shouldn't be worried about parsing command-line input. Instead, I think the keyboard logic should be part of the `net_handler_cli`, maybe running as a thread in there

levincent06: Makes sense to me
So we wouldn't have a `keyboard_interface.c` anymore, but rather it would be a part of `net_handler_cli`?

DanielMolina24: Cant we just have the `keyboard_interface` have its own header then and have the cli include that header and just crate a thread in the cli to execute the keyboard_interface?

benliao1: Sure, we could do that too!

tests/cli/keyboard_interface.c
```diff
@@ -59,63 +21,62 @@ int connect_tcp() {
...
+        memset(joystick_vals, 0, NUM_GAMEPAD_JOYSTICKS * sizeof(float));
+        // Read in bit string of size NUM_GAMEPAD_BUTTONS_AND_JOYSTICKS, not sizeof(gamepad_buf)(NUM_GAMEPAD_BUTTONS_AND_JOYSTICKS + 1) since we are being sent a NUM_GAMEPAD_BUTTONS_AND_JOYSTICKS len string
+        recv(fd, gamepad_buff, NUM_GAMEPAD_BUTTONS_AND_JOYSTICKS, 0);
+        gamepad_buff[NUM_GAMEPAD_BUTTONS_AND_JOYSTICKS] = '\0';
+
+        // Reset keyboard
+        keyboard_buttons = 0;
+        memset(keyboard_buff, 0, NUM_KEYBOARD_BUTTONS * sizeof(char));
+        recv(fd, keyboard_buff, NUM_KEYBOARD_BUTTONS, 0);
+        keyboard_buff[47] = '\0';
```
levincent06, line 75: `keyboard_buff[NUM_KEYBOARD_BUTTONS] = '\0';`

levincent06 (changes requested):
Just one small fix

tests/cli/keyboard_interface.c
```diff
@@ -148,15 +109,22 @@ int main() {
             }
         }
 
-        // Parse the buttons TODO: EXTEND TO BUTTON_XBOX
-        for (int i = button_a; i <= button_xbox; i++) {
-            if (buff[i] == '1') {
-                buttons |= (1 << i);
+        // Set bitmap for gamepad
+        for (int i = 0; i <= NUM_GAMEPAD_BUTTONS; i++) {
```
levincent06, line 113: I think this is a typo `<` vs `<=`

levincent06 (changes requested):

levincent06 (approved):

tests/cli/keyboard_inputs.py
```diff
@@ -114,18 +174,23 @@ def write_to_socket():
...
         mutex.acquire()
-        string_to_send = ''.join(bits)
-        print("sending", string_to_send.encode())
-        sock.send(string_to_send.encode()) # send the 'bitstring' over tcp using socket object
+        gamepad_to_send = ''.join(gamepad_bits)
+        print("sending gamepad", gamepad_to_send.encode())
+        sock.send(gamepad_to_send.encode()) # send the 'bitstring' over tcp using socket object
+        keyboard_to_send = ''.join(keyboard_bits)
+        print("sending keyboard", keyboard_to_send.encode())
+        sock.send(keyboard_to_send.encode())
```
AVSurfer123, line 182: Think it would be easier to send over a 64-bit integer rather than a bit string. But maybe this is fine for now since it works

tests/cli/keyboard_interface.c
AVSurfer123: the convention is in the style guide i think

tests/cli/keyboard_inputs.py
AVSurfer123: Can use `ctypes` to get 64bit int. See https://stackoverflow.com/questions/19716315/how-to-create-a-fixed-size-unsigned-integer-in-python

DanielMolina24: Great suggestion! I'll make a separate issue that can be tackled on a different branch

PR #167, opened 2021-01-15
[EXECUTOR] Fix regex for subscription parsing in student code

### LINKED ISSUE #159: [EXECUTOR] Fix regex in student code matching to detect multiple subscriptions per line
The current regex usage matches only one subscription per line. For example,
```
if Robot.get_value('RED_BOOL') and Robot.get_value('FLIP_FLOP'):
```
will subscribe to only `RED_BOOL`. 

Since the student code needs data about `FLIP_FLOP`, `executor` should be able to match all `Robot.get_value()` function calls.

levincent06:
The regex matching is made to be more precise to capture subscribed params via `Robot.get_value()`. The regex now properly handles whitespace, multiple subscriptions per line, and variable name-sharing.

Important note: if the student code has a variable that shares the name with a variable used in `studentcode.pyx`, then
 `eval(<that variable>)` would cause undefined behavior. For example, if the student named a variable `code_file` in their code, it may evaluate to `studentcode`, as it does in `studentcode.pyx`. To reduce the occurrence of this, we prefixed all variables in `get_all_params` with an underscore, assuming that students won't do the same.

Test `tc_159_1` covers several cases that should (or shouldn't) prompt a subscription. This test may fail for the current executor.
To properly check the device subscriptions, a new assertion `check_sub_requests` was added to `test.c`.

A utility function `check_start_pos()` was also added to `test.c`/`test.h` and `tc_71_1` was changed accordingly to avoid using the shared memory print function.

Finally, this PR also adds additional sleeps to `tc_150_1`  because it sometimes fails due to latency between when a user input is sent and when it changes in shared memory.

Resolves #159


tests/test.c
```diff
@@ -527,6 +527,68 @@ void check_run_mode(robot_desc_val_t expected_run_mode) {
...
+    // Get the shm idx of the specified device
+    dev_id_t dev_ids[MAX_DEVICES];
+    uint32_t catalog;
+
+    get_device_identifiers(dev_ids);
+    get_catalog(&catalog);
+    int shm_idx = -1;
+    if (catalog != 0) {
+        // Iterate through devices
+        for (int i = 0; i < MAX_DEVICES; i++) {
```
AVSurfer123, line 562: You should instead use the `get_dev_ix_from_uid` from `shm_wrapper`. Currently its `static`, but you can make it public and add it to the header file.

AVSurfer123: I'll add it since I'm already working on the branch.

AVSurfer123 (approved):

PR #168, opened 2021-01-15
[MISC] Display help for runtime script

### LINKED ISSUE #163: [MISC] Add brief description and common usages for the Runtime script
When running the Runtime script with no arguments, the following is printed to the terminal:
```
Runtime script needs another argument from:

build run test format clean flash update

```
It will be helpful to briefly describe what each of these commands do and why/how we may want to run it at a glance.
For example, we should mention that `./runtime format -ir .` will Clang format all files in the repo and save it. 

These commands may use the same specific arguments 95% of the time and we shouldn't make the user read through all the possible arguments to figure out the normal use case.

levincent06:
Small additions to the runtime script to display a brief description and common usage for each command.

Closes #163

### DIFF
--- a/runtime
+++ b/runtime
@@ -25,20 +25,6 @@ case $1 in
         popd
         ;;
 
-    run)
-        scripts/run.sh
-        ;;
-
-    test)
-        shift 1
-        scripts/test.sh $@
-        ;;
-
-    format)
-        shift 1
-        scripts/run_clang_format.py $@
-        ;;
-
     clean)
         pushd shm_wrapper && make clean && popd
         pushd net_handler && make clean && popd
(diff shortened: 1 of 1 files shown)


runtime
```diff
@@ -25,10 +25,25 @@ case $1 in
...
+        pushd shm_wrapper && make clean && popd
+        pushd net_handler && make clean && popd
+        pushd executor    && make clean && popd
+        pushd dev_handler && make clean && popd
+        pushd tests       && make clean && popd
+        rm -f /tmp/log-fifo
+        rm -f /tmp/challenge.sock
+        rm -f /tmp/ttyACM*
+        ;;
+
```
DanielMolina24, line 38: Out of curiosity, why the reorder? It seems the only thing that changed right here was the fact you moved it higher in the code

benliao1: Alphabetical order I think? @DanielMolina24

benliao1 (approved):
Looks good!

Will definitely be helpful for the future~~

runtime
DanielMolina24: Oh ok I see!

levincent06: Oh is it in alphabetical order? :P 

I just reordered it so the similar ones were closer to together (ex: build and clean; run and stop)

PR #172, opened 2021-01-23
[SHM] Add keyboard window to the dashboard

### LINKED ISSUE #155: [SHM] Remove print utility functions

### LINKED ISSUE #162: [SHM] Add window for keyboard inputs
With the addition of keyboard inputs for Spring 2021 comp, we should have a window displaying keyboard inputs in addition to gamepad inputs.

levincent06:
The dashboard now has a window to display keyboard inputs (in addition to gamepad inputs).

Other minor changes:
- Made the default IP address of `keyboard_inputs.py` as `127.0.0.1`.
- Print the bitmap being sent by `keyboard_inputs.py` only if it's nonzero.
- Remove `usleep()` in `net_handler_client`. This is needed because `net_handler_cli` uses this to send inputs in realtime.
- `keyboard_interface.c` should be bit-shifting a `1L` instead of `1` because keyboard input bitmap is more than 32 bits.

Resolves #162 
Resolves #155


tests/client/net_handler_client.c
```diff
@@ -448,7 +448,7 @@ void send_user_input(uint64_t buttons, float joystick_vals[4], robot_desc_field_
     free(input.axes);
     free(inputs.inputs);
     free(send_buf);
-    usleep(400000);  // allow time for net handler and runtime to react and generate output before returning to client
+    // usleep(400000);  // allow time for net handler and runtime to react and generate output before returning to client
```
DanielMolina24, line 451: Could we just remove this line instead of commenting it out?

DanielMolina24 (approved):
Nice!

PR #181, opened 2021-02-01
[LOWCAR] Working and tuned KoalaBear PID control and acceleration

benliao1:
Provides for additional adjustability and fine-tuning of the performance of KoalaBears. Several features are added:
- An acceleration constant, which allows us to control how fast a robot can get up to speed (which also dictates how much "inertia" the robot has when slowing down)
- A more accurate calculation of the derivative term of the PID control loop, using a linear regression of a few of the most recent samples of the motor state
- A more thorough attempt at tuning the PID controller coefficients that were determined empirically without tools on staff curry and with the motors free-spinning (in an attempt to make the coefficients usable in many situations)

This also includes a few bug fixes in lowcar:
- Fixed "jumping" when first plugging in the KoalaBear by adding in the variable `online` as a member of the `KoalaBear` class
- Fixed race conditions caused by writing to `enc_a` and `enc_b` while the motors are moving by detaching the interrupts while writing, and then reattaching again once it is done writing
- Fixed `lowcar_print_float` so that it prints negative numbers correctly (it used to print something like `-1.5` as `-1.-500`)

Todos:
- Ensure that the PID coefficients work on not just my KoalaBear
- Ensure that the PID coefficients work on more than just staff curry and free-spinning wheels
- Write a Wiki page explaining PID control and what some of the math does

### DIFF
--- a/lowcar/devices/Device/Messenger.cpp
+++ b/lowcar/devices/Device/Messenger.cpp
@@ -144,18 +144,18 @@ void Messenger::lowcar_printf(char* format, ...) {
 }
 
 void Messenger::lowcar_print_float(char* name, float val) {
-	char sign = (val < 0.0) ? '-' : ' ';				 // Holds sign of the value
-	if (val < 0.0) {
-		val *= -1.0;
-	}
+    char sign = (val < 0.0) ? '-' : ' ';  // Holds sign of the value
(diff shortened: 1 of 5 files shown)


DanielMolina24 (approved):
Awesome job!

PR #183, opened 2021-02-04
[SYSTEMD] Add service file to startup NGROK using systemd

DanielMolina24:
- Service file added to `systemd` folder with the necessary requirements and instructions on when it should start up
- A `.yml` file was also added in `systemd`where the authtoken must be added
- The service file also specifies the path that NGROK should be located in to work


systemd/ngrok.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Ngrok startup process for Runtime
+After=network.target net_handler.service
```
levincent06, line 3: Does this work without the `network.target`?

levincent06:
Can you add to the repo README on how and where to install ngrok?

AVSurfer123:
Well I wouldn't say it is actually part of runtime overall. No one other than the competition robots will need `ngrok` installed. And hopefully after this semester we will never touch `ngrok` again.

systemd/ngrok.service
AVSurfer123: Yea I think this requisite can/should be removed.

AVSurfer123 (approved):

PR #185, opened 2021-02-07
[NET_HANDLER] readonly attribute added to Param protos

### LINKED ISSUE #152: [NET_HANDLER] Include readability/writeability of params in device data
To make it easier for students to get the readability and writeability of params, Dawn should be able to display all of the params, their properties, and values in its peripheral window. We should add the readability and writeability of params to the current device data protobuf, and add to the UDP connection code to forward that information to Dawn.

AVSurfer123:
Adding flag to the `Param` protobuf whether the param is read only for the student.
Also changed how UDP data is tested by directly checking the `DevData*` struct rather than doing string matching.

Closes #152


benliao1:
Looks good, the only thing I'd add is maybe a test to make sure that the info is sent to Dawn?

AVSurfer123:
TODO: Need to add assertion functions to `test.c` that will check the provided `DevData*` struct for correctness. This is since I removed the printing from the `net_handler_client` and moved it into the `net_handler_cli` where it belongs. I currently don't have the time to do this though if someone else wants to finish it.

tests/test.c
```diff
@@ -489,6 +489,42 @@ void check_inputs(uint64_t expected_buttons, float expected_joysticks[4], robot_
...
+        print_fail();
+        fprintf_delimiter(stderr, "Expected device with type %u uid %llu at index %d but got type %u uid %llu instead",
+                          type, uid, index, device->type, device->uid);
+        end_test();
+        exit(1);
+    }
+    print_pass();
+}
+
+void check_udp_device_param(DevData* dev_data, int dev_idx, char* param_name, param_type_t param_type, param_val_t param_val, uint8_t readonly) {
```
levincent06, line 512: Is this function done? What is it used for? Should this be added to the header file?

tests/test.h
```diff
@@ -83,6 +83,20 @@ void add_unordered_string_output(char* output);
  */
 void check_inputs(uint64_t expected_buttons, float expected_joysticks[4], robot_desc_field_t source);
 
+/******************** UDP Device Data Checks ******************/
+
+/**
+ * Checks that the given device was returned by the UDP thread.
```
levincent06, line 89: Wdym by returned by the UDP thread? All device data are sent over UDP right?

tests/test.h
```diff
@@ -83,6 +83,20 @@ void add_unordered_string_output(char* output);
...
+ * 
+ * Args:
+ *  dev_data: Protobuf struct that is outputted by the UDP thread from get_next_device_data()
+ *  index: index where device should be
```
levincent06, line 93: How does a test know which index a particular device would be in the protobuf?

tests/test.h
```diff
@@ -83,6 +83,20 @@ void add_unordered_string_output(char* output);
...
+ *  type: what type the device should be
+ *  uid: what uid the device should be
+ * 
+ */
+void check_udp_device_exists(DevData* dev_data, int index, uint8_t type, uint64_t uid);
```
levincent06, line 98: Would it make more sense to name this `check_device_data_sent()` or similar?

tests/client/net_handler_client.c
```diff
@@ -302,11 +283,9 @@ void start_net_handler() {
         null_fp = fopen("/dev/null", "w");
 
         // init the mutex that will control whether udp prints to screen
```
levincent06, line 285: "init the lock on the global variable that holds the most recent device data"?

tests/client/net_handler_client.h
```diff
@@ -82,11 +82,10 @@ void send_challenge_data(robot_desc_field_t client, char** data, int num_challen
 void send_device_subs(dev_subs_t* subs, int num_devices);
 
 /**
- * Calling this function will let the next device data packet coming into Dawn from Runtime
- * to be printed to standard output (since device data packets are constantly coming in, they
- * are normally suppressed; this function unsuppresses one single packet).
+ * Calling this function will return the most recent device data packet coming into Dawn from Runtime.
+ * NOTE: you must free the pointer returned by this function using dev_data__unpacked
```
levincent06, line 86: Do you mean `dev_data__free_unpacked()`?

tests/client/net_handler_client.c
```diff
@@ -4,18 +4,20 @@
...
 
 int nh_tcp_shep_fd = -1;      // holds file descriptor for TCP Shepherd socket
 int nh_tcp_dawn_fd = -1;      // holds file descriptor for TCP Dawn socket
 int nh_udp_fd = -1;           // holds file descriptor for UDP Dawn socket
 FILE* default_tcp_fp = NULL;  // holds default output location of incoming TCP messages (stdout if NULL)
 FILE* tcp_output_fp = NULL;   // holds current output location of incoming TCP messages
-FILE* udp_output_fp = NULL;   // holds current output location of incoming UDP messages
 FILE* null_fp = NULL;         // file pointer to /dev/null
 
+DevData* device_data = NULL;
```
levincent06, line 15: Maybe add a comment that this holds the most recent device data received

levincent06 (changes requested):

tests/cli/net_handler_cli.c
```diff
@@ -315,6 +315,35 @@ void prompt_reroute_output() {
     update_tcp_output_fp(dest);
 }
 
+void print_next_dev_data() {
```
DanielMolina24, line 318: Can you explain why you decided to move this function to the net_handler_cli? Personally, I always viewed the cli as  an area that displays values or takes in inputs, rather than getting the data and parsing it. I feel the majority of this should be within the client

tests/test.c
AVSurfer123: I've finished this function now

tests/test.h
AVSurfer123: I meant sent by the UDP thread. I will fix it

tests/test.h
AVSurfer123: They probably just need to run it out on their own using the CLI. it's also the same order as device indices in SHM, but with any gaps removed.

tests/test.h
AVSurfer123: Well its already sent, but we are checking if a specific device was sent in that packet.

tests/cli/net_handler_cli.c
AVSurfer123: Talked about this during a meeting some time ago.

net_handler/udp_conn.c
```diff
@@ -94,6 +94,7 @@ static void* send_device_data(void* args) {
                     }
                     param__init(param);
                     param->name = device_info->params[j].name;
+                    param->readonly = device_info->params[j].write == 0;
```
benliao1, line 97: Can we add a TODO comment or something after this to make sure we know to use `stdbool.h` once we get around to using that header? And to make it more explicit, I usually prefer writing it as
```
device_info->params[j].write ? 0 : 1;
```

tests/client/net_handler_client.c
```diff
@@ -622,6 +600,17 @@ void send_device_subs(dev_subs_t* subs, int num_devices) {
     usleep(400000);  // allow time for net handler and runtime to react and generate output before returning to client
 }
 
+DevData* get_next_dev_data() {
```
benliao1, line 603: Can you add some comments on and in this function to explain what it does?

benliao1 (commented):
Just a couple of small comments!

net_handler/udp_conn.c
AVSurfer123: I made it check for `false` now. I also don't like using the ternary operator for this.

benliao1: I think it's more explicit; also if you don't specifically set it to 1, the true value could be any nonzero value, which can cause problems up the line when packing the protobuf / interpreting the boolean value.

AVSurfer123: The C language will only return 1 or 0 for an == operation, so you don't need to worry about that.

tests/client/net_handler_client.c
```diff
@@ -80,62 +82,38 @@ static int connect_tcp(robot_desc_field_t client) {
 
 /**
  * Function to receive data from net_handler on the UDP port
```
levincent06, line 84: Add a comment that it updates the global variable with the incoming DeviceData (Rn from the header it's hard to tell what the function output is)

tests/integration/tc_71_18.c
```diff
@@ -22,24 +22,33 @@ char dev2_header[] = "Device No. 1:\ttype = SimpleTestDevice, uid = 17185, itype
...
+char* parameters[NUM_PARAMS] = {
+    increasing,
+    doubling,
+    flipflop,
+    myint,
+};
 
 /**
  * Helper function to send dev1_params subscriptions
  * to device 1 and dev2_params subscriptions to device 2
```
levincent06, line 41: Add a comment mentioning that it returns a DeviceData message that *should* reflect the new subscription, and that this message should be checked

levincent06 (changes requested):

tests/client/net_handler_client.c
AVSurfer123: Fixed

tests/integration/tc_71_18.c
AVSurfer123: Added

levincent06 (approved):

PR #192, opened 2021-02-23
[DOCS] Added Sphinx docs autogeneration

AVSurfer123:
Sphinx docs for the Student API. The theme and look should probably be changed in the future


AVSurfer123:
@benliao1 we can merge this in for now, it won't harm us to have it and it can always be changed later.

benliao1:
This would not have merged cleanly into master I think, so I just merged master into it and added the submodule. I'll approve it once it passes the build.

benliao1 (approved):
Looks good, can be updated / changed later if we want to make our student API documentation better!

PR #194, opened 2021-03-18
[MISC] 2021 Dawn Integration Changes

benliao1:
This PR makes a few bugfixes to get Runtime to work fully with the 2021 version of Dawn. It also adds an additional service to automatically start the `ngrok` process to expose three necessary ports to the Internet and display their IP addresses on a port accessible within the local network the Raspberry Pi is connected to.


AVSurfer123 (approved):

tests/cli/keyboard_interface.c
```diff
@@ -146,7 +146,7 @@ void setup_keyboard() {
         // Set bitmap for keyboard
         for (int i = 0; i < NUM_KEYBOARD_BUTTONS; i++) {
             if (keyboard_buff[i] == '1') {
-                keyboard_buttons |= (1L << i);
+                keyboard_buttons |= (1LL << i);
```
levincent06, line 149: Let's typecast the 1
I think it would be more clear and easy to understand at a glance.

tests/cli/shm_ui.c
```diff
@@ -356,7 +356,7 @@ void display_keyboard_state(char** key_names) {
         wclrtoeol(KEYBOARD_WIN);
         // Show button name; If pressed, make it bold. Else, dim
         if (keyboard_connected) {
-            if (pressed_buttons & (1L << i)) {
+            if (pressed_buttons & (1LL << i)) {
```
levincent06, line 359: Same comment as above

systemd/ngrok_displayer.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
```
levincent06, line 1: If we plan on using ngrok in the future (ex: more remote comps(?)) I think it would be a good idea to leave a comment here explaining what this script is for and how it should be used

levincent06 (changes requested):

DanielMolina24 (approved):

levincent06 (approved):

PR #196, opened 2021-03-21
[NET_HANDLER][DEV_HANDLER][CLI] Add CLI attaching

### LINKED ISSUE #118: [CLI] Make CLIs attach to existing instances of Runtime
When starting up each of the CLIs, they each spawn the corresponding instance of the Runtime component.
(Starting up net hander cli will `fork()`, then the child process will `exec()` a new net handler process.)

We would rather require that net handler is already started, then net handler CLI will simply feed commands to that process. This will prevent issues involving duplicate instances of Runtime, which may happen when the CLIs crash for whatever reason. This applies to all the main components of Runtime (shared memory, net handler, executor, and dev handler).

This will require changes to the `start()` functions in the clients. This will also affect the testing framework, as each individual test requires that specific components of Runtime are spawned at the start then cleaned up at the end.

levincent06:
Net handler CLI and dev handler CLI can now "attach" to their respective existing instances of runtime instead of spawning a new instance.
- Net handler CLI can specify zero or more flags from (`-d` and `-s`)
  - Without arguments it spawns a new net handler instance and connects a fake Dawn and fake Shepherd.
    - This is unchanged and the same behavior as on `master`.
  - `-d` : Does not spawn a new net handler; Connects a fake Dawn
  - `-s` : Does not spawn a new net handler; Connects a fake Shepherd
- Dev handler no longer takes in arguments.
  - Dev handler now detects BOTH Arduinos and virtual devices.
 - Dev handler CLI has an optional argument `attach`.
   - Without this argument, it spawns a new dev handler. This is unchanged and the same behavior as on `master`.
   - With this argument, it **doesn't** spawn a new dev handler. Disconnecting/connecting virtual devices will work normally with an existing instance of dev handler.

Attempting to use these CLIs in their respective "attach" modes without a preexisting instance of runtime is will have undefined behavior.

Closes #118

### DIFF
--- a/tests/cli/net_handler_cli.c
+++ b/tests/cli/net_handler_cli.c
@@ -15,9 +15,9 @@
  * If no flags are specified, both (fake) Dawn and (fake) Shepherd will be connected to a new
  * net handler instance.
  */
-int spawned_net_handler = 1;  // Whether the CLI spawned a new net handler instance
-int dawn = 0;                 // Whether to connect a fake Dawn
-int shepherd = 0;             // Whether to connect a fake Shepherd
+bool spawned_net_handler = true;  // Whether the CLI spawned a new net handler instance
+bool dawn = false;                // Whether to connect a fake Dawn
+bool shepherd = false;            // Whether to connect a fake Shepherd
 
 // ********************************** COMMAND-SPECIFIC FUNCTIONS  ****************************** //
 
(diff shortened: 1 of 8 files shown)


tests/cli/net_handler_cli.c
```diff
@@ -345,21 +360,51 @@ void connect_keyboard() {
...
     signal(SIGINT, sigint_handler);
 
+    // Parse command line flags
+    int c;
+    opterr = 0;
+    while ((c = getopt(argc, argv, "ds")) != -1) {
+        switch (c) {
+            case 'd':
+                spawned_net_handler = 0;
+                dawn = 1;
```
benliao1, line 376: Should these variables be of the type `bool` that you defined in `runtime_util`? I didn't realize you could do that, we should change all of our flag variables to this type, kinda like a "pseudo-boolean" type. We should make sure that the `bool` type  you defined takes on the underlying type of `uint8_t` though. Not sure how to do that.

benliao1 (commented):
Great work! Will this require modification to the tests at all?

benliao1 (approved):
Looks good! Linking issue #198 here to remind ourselves to eventually transition to using that type everywhere in Runtime.

PR #200, opened 2021-03-24
[MISC] Implement ngrok restarting from webpage

### LINKED ISSUE #197: [MISC] Add button to generate new ngrok session
For Spring 2021, each time a new team wants to control the robot, we need to generate a new set of ngrok IP addresses. (This is to prevent different teams from connecting to Runtime simultaneously.)

The current way to do this is to `ssh` into the Pi and run `sudo systemctl restart ngrok`. This will run `ngrok.sh`, which restarts ngrok with new IP addresses and displays them on a webpage (through `ngrok_displayer.py`). 

A new, more user-friendly solution, is to add an HTML button on the webpage which will restart ngrok and display the new IP addresses.

levincent06:
- `ngrok_displayer` is its own service.
- Accessing `/new` endpoint will restart `ngrok.service` first.
- Refreshing webpage will fetch the current IP addresses before
  displaying anything.

Resolves #197

### DIFF
--- a/systemd/ngrok.sh
+++ b/systemd/ngrok.sh
@@ -2,14 +2,8 @@
 
 function sigint_handler {
     pkill -u pi /home/pi/ngrok
-    pkill -u pi ngrok_displayer
 }
 
 trap 'sigint_handler' INT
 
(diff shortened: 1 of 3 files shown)


benliao1 (approved):
Changes look good and make sense! I tested it in 101 and it works, so will push changes to all six raspberry pi's.

PR #201, opened 2021-03-28
[LOWCAR] Several lowcar fixes

levincent06:
- The Device Type is a 1 byte value. This is already the case in the Runtime code, but it seems like we forgot to change this in the lowcar code.
- "Enabled" in lowcar is clarified to mean "connected to Runtime". In the constructor, `enabled` is false. When `enabled` is set to true (i.e. when the device sends an `ACK`), we call `device_enable()`.

### DIFF
--- a/scripts/flash.sh
+++ b/scripts/flash.sh
@@ -52,8 +52,6 @@ function usage {
         printf "\t$type\n"
     done
     printf "\n"
-    
-    printf "Remember that you need to install arduino-cli first before you can use this flash script!\n\n"
     exit 1
 }
 
@@ -312,18 +310,17 @@ arduino-cli config init
 # update the core index (see arduino-cli documentation)
 arduino-cli core update-index
 
+# symlink libraries
+symlink_libs
+
+# insert uid and requested device into Device_template, copy into Device/Device.ino
+make_device_ino
+
 # get the device port, board, fqbn, and core
 get_board_info
 
 # install this device's core
 arduino-cli core install $DEVICE_CORE
-
-# symlink libraries
-get_lib_install_dir
-symlink_libs
-
-# insert uid and requested device into Device_template, copy into Device/Device.ino
-make_device_ino
 
 # compile the code
 printf "\nCompiling code...\n\n"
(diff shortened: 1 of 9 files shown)


benliao1 (approved):
Makes sense! Will need to test the `enable` for specifically the motor controllers and the servo controller, since those have more than just pin mode sets in them.

AVSurfer123:
Added 2 more commits to this PR from the `ping_rename` cus I was lazy
- Renamed `PING` to `DEVICE_PING` to avoid name conflicts with an Arduino Mega
- Reordered functions in `flash.sh` so that `Device` folder gets created regardless

scripts/flash.sh
```diff
@@ -52,8 +52,6 @@ function usage {
         printf "\t$type\n"
     done
     printf "\n"
-    
-    printf "Remember that you need to install arduino-cli first before you can use this flash script!\n\n"
     exit 1
```
benliao1, line 55: Why did we delete this message?

scripts/flash.sh
```diff
@@ -312,19 +310,18 @@ arduino-cli config init
...
 arduino-cli core update-index
 
-# get the device port, board, fqbn, and core
-get_board_info
-
-# install this device's core
-arduino-cli core install $DEVICE_CORE
-
 # symlink libraries
-get_lib_install_dir
```
benliao1, line 322: Why did we change the flash script? Don't you need to determine what directory to install the libraries before symlinking them? They are different on different Linux distros and on Macs, so this function is important!

benliao1 (commented):
A couple questions about the modifications to the flash script, I think @AVSurfer123 made these changes? I'm just not sure if we should be making these changes~~

scripts/flash.sh
AVSurfer123: It is already printed at the bottom if the script can't find `arduino-cli` on the path

scripts/flash.sh
AVSurfer123: I guess its a bit weird, but this is already called in the `clean` function, and we always `clean` before doing anything else so no need to call it twice.

benliao1: Ah okay, yea then this is I guess a question of whether we should `clean` before doing anything each time, but for now this makes sense to delete this.

scripts/flash.sh
benliao1: Oh okay I didn't realize that, that makes sense!

PR #204, opened 2021-04-01
[NET_HANDLER] Implemented issue #203

### LINKED ISSUE #203: [NET_HANDLER] Block dawn enter_auto and enter_teleop with Shepherd connected
Small feature, probably like five lines of code in `net_handler`. If Shepherd is connected to the robot, Dawn should not be able to change the run mode of the robot to `AUTO` or `TELEOP` (Dawn can enter idle though i.e. the student can E-Stop the robot while Shepherd is connected if they want.)

This is to prevent students from switching the run mode of the robot while Shepherd is connected (when the robot is running a match).

We should write a test case to test this behavior though!

benliao1:
- Added a long if statement in `tcp_conn.c` to implement desired behavior
- Also disconnected keyboard on Dawn disconnection

Closes #203

### DIFF
--- a/net_handler/tcp_conn.c
+++ b/net_handler/tcp_conn.c
@@ -184,8 +184,9 @@ static int recv_new_msg(int conn_fd, int challenge_fd) {
             return -2;
         }
 
-        // if shepherd is connected and dawn tries to send RUN_MODE == AUTO or TELEOP, block it
-        if (pthread_self() == dawn_tid && robot_desc_read(SHEPHERD) == CONNECTED && (run_mode_msg->mode == MODE__AUTO || (run_mode_msg->mode == MODE__TELEOP))) {
+        //if shepherd is connected and dawn tries to set RUN_MODE == AUTO or TELEOP, block it
+        if (pthread_self() == dawn_tid && robot_desc_read(SHEPHERD) == CONNECTED &&
+            (run_mode_msg->mode == MODE__AUTO || (run_mode_msg->mode == MODE__TELEOP))) {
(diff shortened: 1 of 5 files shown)


net_handler/tcp_conn.c
```diff
@@ -183,6 +184,12 @@ static int recv_new_msg(int conn_fd, int challenge_fd) {
             return -2;
         }
 
+        // if shepherd is connected and dawn tries to send RUN_MODE == AUTO or TELEOP, block it
+        if (pthread_self() == dawn_tid && robot_desc_read(SHEPHERD) == CONNECTED && (run_mode_msg->mode == MODE__AUTO || (run_mode_msg->mode == MODE__TELEOP))) {
+            log_printf(INFO, "You cannot send Robot to Auto or Teleop from Dawn with Shepherd connected!");
```
DanielMolina24, line 189: Very minor Q, but did you mean to write "set" instead of "send"?

levincent06 (approved):

net_handler/tcp_conn.c
benliao1: I meant to write "send" but you're right, "set" make more sense.

benliao1:
This changes makes it impossible to test whether Dawn can change to `TELEOP` and `AUTO` when Shepherd is not connected. This behavior is pretty important to verify, so we should probably change the tests clients to be able to accommodate the changes in #196 for `net_handler_client`. In other words, there should be an option to disconnect one or both Dawn and Shepherd in `net_handler_client.h`.

PR #205, opened 2021-04-01
[NET_HANDLER] Implemented Latency Protobuf

### LINKED ISSUE #195: [NET_HANDLER] Implement latency check with Dawn
Dawn should have the functionality to check the latency between Dawn and Runtime (more specifically, net handler).
One way I could see this implemented is that when Dawn wants to check latency (either by polling or by pressing a button on the GUI):
1. Dawn sends a protobuf with the current timestamp.
2. Net handler receives it, appends the current timestamp, and sends it back to Dawn.
3. Dawn receives the protobuf and determines the latency: `<current timestamp> - <timestamp in step 1>`

The timestamp appended by runtime in step 2 may be useful in the future, but we should be careful about clocks being not synchronized across devices.

DanielMolina24:
- Added a new TimeStamp proto with two fields, `dawn_timestamp` and `runtime_timestamp`, both Fixed64
- Added handling in `tcp_conn` where we simply append a new timestamp to the TimeStamp message we receive
- Implemented a way to test it in `net_handler_client` that emulates how Dawn sends Runtime TimeStamp protos

Closes #195


tests/client/net_handler_client.c
```diff
@@ -163,26 +163,38 @@ static int recv_tcp_data(robot_desc_field_t client, int tcp_fd) {
...
+            for (int i = 0; i < msg->n_payload; i++) {
+                fprintf(tcp_output_fp, "%s", msg->payload[i]);
+            }
+        } else if (msg_type == CHALLENGE_DATA_MSG) {
+            for (int i = 0; i < msg->n_payload; i++) {
+                fprintf(tcp_output_fp, "Challenge %d result: %s\n", i, msg->payload[i]);
+            }
         }
+        fflush(tcp_output_fp);
+        text__free_unpacked(msg, NULL);
```
levincent06, line 193: Seems like the `if` cases are incorrect. Is the following what was intended?
```
if (msg_type == TIME_STAMP_MSG) {
...
} else if (msg_type == LOG_MSG) {
...
} else if (msg_type == CHALLENGE_DATA_MSG) {
...
} else {
// Unhandled case
}
```

levincent06 (changes requested):

tests/client/net_handler_client.c
benliao1: This would probably make more sense; in the past we would only be receiving text messages from Runtime (`LOG` and `CHALLENGE_DATA`) so I had just a simple unpack as text protobuf and then depending on the message type, print it out a certain way. But now that we're receiving three different types of messages, might be easier to do `if` - `else-if` - `else` like you suggested above.

benliao1 (commented):
Looks good! I think we just have to make a decision about Vincent's suggestion and then I had a couple of really picky things and we should be good~~

tests/client/net_handler_client.c
DanielMolina24: Its not that the if statements are wrong, its more so I just grouped all the ones relating to text unpacking into one condition and then had additional checks in that one condition to see the msg_type. Functionality wise it's still the same as before. I made it that way to just have one place where text is unpacked then we can see which condition to enter.

DanielMolina24: I can modify it to fit the `if`-`else if`- `else` structure if you want it to be cleaner though

net_handler/tcp_conn.c
```diff
@@ -107,6 +107,23 @@ static void send_log_msg(int conn_fd, FILE* log_file) {
     free(send_buf);
 }
 
+/*
+* Send a timestamp message over TCP with a timestamp attached to the message. The timestamp is when the net_handler on runtime has received the
+* packet and processed it. 
+* Arguments:
+    - int conn_fd: socket connection's file descriptor on which to write to the TCP port
+    - TimeStamps* dawn_timestamp_msg: Unpacked timestamp_protot from Dawn
```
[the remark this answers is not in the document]
DanielMolina24: lol I cant spell

tests/client/net_handler_client.c
```diff
@@ -163,26 +163,38 @@ static int recv_tcp_data(robot_desc_field_t client, int tcp_fd) {
...
-    // print the incoming message
-    if (msg_type == LOG_MSG) {
-        for (int i = 0; i < msg->n_payload; i++) {
-            fprintf(tcp_output_fp, "%s", msg->payload[i]);
+    if (msg_type == TIME_STAMP_MSG) {
+        TimeStamps* time_stamp_msg = time_stamps__unpack(NULL, len, buf);
+        if (time_stamp_msg == NULL) {
+            log_printf(ERROR, "recv_new_msg: Cannot unpack time_stamp msg");
+        }
+        uint64_t final_timestamp = millis();
```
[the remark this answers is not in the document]
DanielMolina24: So `mills()` in runtime_util.c calls `gettimeofday`which gives us the number of seconds and microseconds since epoch. We then do some manipulation to turn that into milliseconds and send it on its way to Dawn. It seems to work from what we tested today

tests/client/net_handler_client.c
levincent06: Oh I see! I think it's just more intuitive to me if it was a `switch`/`case` altogether with a case for each message type, but this is fine!

levincent06 (approved):

tests/client/net_handler_client.c
DanielMolina24: Yea I do see your point and agree, I'll change it and push it in a bit

tests/client/net_handler_client.c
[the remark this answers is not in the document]
benliao1: Ah okay I didn't realize `millis()` was our own function, I think wrote that sometime in the summer of last year haha
But ok sounds good, glad to know it works!

benliao1:
Removed `net_handler/pbc_gen` from git because those are generated files. Made the `net_handler` `Makefile` make the directory if it doesn't exist, to put all the compiled protobufs in there.

benliao1:
ffc2424 deleted stuff relating to removing the `pbc_gen` folder from the repo, so I cherry-picked the commit from my local branch and force pushed it after rebasing off of master to make it up-to-date with the base branch.

net_handler/tcp_conn.c
```diff
@@ -283,7 +298,15 @@ static int recv_new_msg(int conn_fd, int challenge_fd) {
                 break;
             default:
                 log_printf(ERROR, "requested gamestate to enter invalid state %s", game_state_msg->state);
+                game_state__free_unpacked(game_state_msg, NULL);
+        }
+    } else if (msg_type == TIME_STAMP_MSG) {
+        TimeStamps* time_stamp_msg = time_stamps__unpack(NULL, len_pb, buf);
+        if (time_stamp_msg == NULL) {
+            log_printf(ERROR, "recv_new_msg: Cannot unpack time_stamp msg");
```
AVSurfer123, line 306: should put a `return` here

PR #210, opened 2021-04-03
[SHM][UI] Add check in shm_init() that shm exists

levincent06:
- UI fails in attach mode if shm doesn't exist.
- UI polls for the existence of shm and will exit if it was destroyed.
- Fix minor visual bug with the note at the bottom of the UI not displaying properly.


DanielMolina24 (approved):
Looks good!

PR #211, opened 2021-04-04
[EXECUTOR][NET_HANDLER] Changed log level in kill_subprocess()

DanielMolina24:
- signal handler for executor commented out
- log level dependent on signal received
- new protocol implemented for net_handler:
- Essentially determine the client, with a timeout if client ID isn't sent during a set amount of time
- Kill the previous version of dawn or shepherd and form a new connection with what is likely a new instance of Dawn and Shepherd

### DIFF
--- a/executor/executor.c
+++ b/executor/executor.c
@@ -513,7 +513,7 @@ static pid_t start_mode_subprocess(char* student_code, char* challenge_code) {
             robot_desc_write(RUN_MODE, IDLE);  // Will tell parent to call kill_subprocess
         } else {
             executor_init(student_code);
-            signal(SIGTERM, python_exit_handler); // kill subprocess regardless
+            signal(SIGTERM, python_exit_handler);  // kill subprocess regardless
             run_mode(mode);
         }
         exit(0);
--- a/executor/executor.c
+++ b/executor/executor.c
@@ -516,7 +516,7 @@ static pid_t start_mode_subprocess(char* student_code, char* challenge_code) {
             robot_desc_write(RUN_MODE, IDLE);  // Will tell parent to call kill_subprocess
         } else {
             executor_init(student_code);
-            signal(SIGTERM, python_exit_handler);
+            // signal(SIGTERM, python_exit_handler); // kill subprocess regardless
             run_mode(mode);
         }
         exit(0);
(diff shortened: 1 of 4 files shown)


benliao1 (approved):
Makes sense to me!

AVSurfer123 (commented):
Why is the point of this? If its because the fix I told Ben to do with commenting out the `python_exit_handler`, then that isn't being included here. Otherwise, you will still want an unhandled SIGTERM to print this out.

levincent06:
The issue is that because executor kills the subprocess with `SIGTERM`, it still prints out
```
Error when shutting down execution of mode <mode>
killed by signal 15
```
every single time the run mode changes from `AUTO`/`TELEOP` even though it's not really an error.

AVSurfer123:
That shouldn't print out as there is a signal handler attached to SIGTERM. The only way it gets print out is if you comment/remove the signal handler, but that isn't part of this PR.

executor/executor.c
```diff
@@ -482,12 +484,9 @@ static void kill_subprocess() {
...
     }
-    if (!WIFEXITED(status)) {
-        log_printf(ERROR, "Error when shutting down execution of mode %d", mode);
-    }
-    if (WIFSIGNALED(status)) {
-        log_printf(ERROR, "killed by signal %d\n", WTERMSIG(status));
-    }
+    int sig_number = WTERMSIG(status);
+    log_printf((sig_number == DEBUG_SIGNAL) ? DEBUG : ERROR, "Error when shutting down execution of mode %d", mode);
+    log_printf((sig_number == DEBUG_SIGNAL) ? DEBUG : ERROR, "killed by signal %d\n", sig_number);
```
AVSurfer123, line 489: Instead of doing this, can you just put `exit(0)` at the beginning of `python_exit_handler`? and I guess comment the rest of the handler for now.

net_handler/net_handler.c
```diff
@@ -66,6 +66,35 @@ static void sigint_handler(int sig_num) {
...
+    int ret = select(connfd + 1, &set, NULL, NULL, &timeout);
+    if (ret == -1) {
+        log_printf(ERROR, "determine_client: Error on select"); /* an error accured */
+        return 255;
+    } else if (ret == 0) {
+        log_printf(ERROR, "determine_client: Timeout on waiting for client ID"); /* a timeout occured */
+        return 254;
+    } else { /* there was data to read */
+        uint8_t client_id = 255;
+        read(connfd, &client_id, 1);
```
AVSurfer123, line 93: check for error on read

AVSurfer123 (changes requested):

executor/executor.c
AVSurfer123: yea. but just comment out the other code in `python_exit_handler` for now.

AVSurfer123 (approved):

PR #222, opened 2021-06-08
[LOWCAR] Added optional path argument to flash script; resolves #202

### LINKED ISSUE #202: [LOWCAR] Add optional path argument to flash script
Currently, the flash script will flash devices located at only `/dev/ttyACM0`. It may be useful to add an optional argument to specify exactly which connected device we want to flash.

benliao1:
Small QOL feature. This is especially useful for Shepherd, when they might have many Arduinos attached to the field control computer at the same time and specifically want to flash one of them. Could be useful for Runtime as well, if we're testing multiple devices on the Raspi and are flashing from it.

Also updated usage text with the new path flag (`-p`), and added list of available ports on computer as part of the help text.

Closes #202

### DIFF
--- a/scripts/flash.sh
+++ b/scripts/flash.sh
@@ -36,15 +36,17 @@ CLEAN=0                   # whether or not the clean option was specified
 
 # prints the usage of this flash script
 function usage {
-    printf "Usage: flash [-cht] [-s UID] device_type\n"
+    printf "Usage: flash [-cht] [-s UID] [-p port] device_type\n"
     printf "\t-c: short for \"Clean\"; removes the \"Device\" folder and all symbolic links\n"
     printf "\t-h: short for \"Help\": displays this help message\n"
+    printf "\t-p: short for \"Port\": manually specify the port file path of the Arduino to flash\n"
     printf "\t-s: short for \"Set\"; manually set the UID for the device (input taken in hex, do not prefix \"0x\")\n"
     printf "\t-t: short for \"Test\"; set the UID for the device to \"0x123456789ABCDEF\"\n\n"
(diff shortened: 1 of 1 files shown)


scripts/flash.sh
```diff
@@ -36,22 +36,48 @@ CLEAN=0                   # whether or not the clean option was specified
...
     printf "\t-c: short for \"Clean\"; removes the \"Device\" folder and all symbolic links\n"
     printf "\t-h: short for \"Help\": displays this help message\n"
+    printf "\t-p: short for \"Port\": manually specify the port file path of the Arduino to flash\n"
     printf "\t-s: short for \"Set\"; manually set the UID for the device (input taken in hex, do not prefix \"0x\")\n"
     printf "\t-t: short for \"Test\"; set the UID for the device to \"0x123456789ABCDEF\"\n\n"
     printf "Examples:\n"
     printf "\tTo flash a DummyDevice: flash DummyDevice\n"
     printf "\tTo flash a DummyDevice specifying the UID \"ABCDEF\": flash -s ABCDEF DummyDevice\n"
     printf "\tTo flash a DummyDevice with the test UID \"123456789ABCDEF\": flash -t DummyDevice\n"
+    printf "\tTo flash a DummyDevice that is at the port \"/dev/ttyACM1\": flash -p /dev/ttyACM1 DummyDevice\n"
```
AVSurfer123, line 49: You don't need to change this, but in the future it might be easier to use single quotes ' ' as the outside quotes so you don't need to escape the double quotes " " inside.

AVSurfer123 (approved):

PR #224, opened 2021-06-23
[EXECUTOR] Removed any references to coding challenges from runtime

DanielMolina24:
Hasta la vista coding challenges


AVSurfer123 (changes requested):

net_handler/tcp_conn.c
```diff
@@ -332,7 +283,7 @@ static void* tcp_process(void* tcp_args) {
     //variables used for waiting for something to do using select()
     fd_set read_set;
     int log_fd;
-    int maxfd = args->challenge_fd > args->conn_fd ? args->challenge_fd : args->conn_fd;
+    int maxfd = args->conn_fd;
```
benliao1, line 286: idk if this makes it easier to understand or not but what do you think about
```
if (args->send_logs) {
    log_fd = fileno(args->log_file);
    maxfd = log_fd > args->conn_fd ? log_fd : conn_fd;
} else {
    maxfd = args->conn_fd;
}
```

net_handler/udp_conn.c
```diff
@@ -94,6 +94,7 @@ static void* send_device_data(void* args) {
                     }
                     param__init(param);
                     param->name = device_info->params[j].name;
+                    param->readonly = device_info->params[j].write == false;
```
benliao1, line 97: is this a change that you made? or is this a merge conflict with the base branch

net_handler/udp_conn.c
```diff
@@ -167,6 +168,7 @@ static void* send_device_data(void* args) {
         time->name = "time_ms";
         time->val_case = PARAM__VAL_IVAL;
         time->ival = millis() - start_time;  // Can only give difference in millisecond since robot start since it is int32, not int64
+        time->readonly = true;
```
benliao1, line 171: same thing as this

tests/cli/net_handler_cli.c
```diff
@@ -330,6 +269,35 @@ void prompt_reroute_output() {
     update_tcp_output_fp(dest);
 }
 
+void print_next_dev_data() {
```
benliao1, line 272: was this function not here before??

tests/client/net_handler_client.c
```diff
@@ -622,6 +565,19 @@ void send_device_subs(dev_subs_t* subs, int num_devices) {
     usleep(400000);  // allow time for net handler and runtime to react and generate output before returning to client
 }
 
+DevData* get_next_dev_data() {
+    pthread_mutex_lock(&device_data_lock);
+    // We need to copy the data in the device_data Protobuf to return to the caller
+    // The easiest way I found was to pack it into a buffer then unpack it again, but this may not be optimal
```
benliao1, line 571: don't know if this an old comment but if we just want a copy of the device data, you can do use the function `memcpy` to copy bytes from one piece of memory to another; see https://www.tutorialspoint.com/c_standard_library/c_function_memcpy.htm

tests/integration/tc_71_8.c
```diff
@@ -9,15 +9,13 @@
...
+    start_test("Hotplug ForeignTestDevice", "", NO_REGEX);
 
     // Connect a ForeignTestDevice
     check_device_not_connected(UID);
     connect_virtual_device("ForeignTestDevice", UID);
     sleep(2);
     check_device_not_connected(UID);
 
-    // Clean up
-    end_test();
```
benliao1, line 21: this might also be old but why are the `end_test()` calls deleted in all of the integration tests?

net_handler/net_util.h
```diff
@@ -40,7 +41,6 @@
 typedef enum net_msg {
     RUN_MODE_MSG,
     START_POS_MSG,
-    CHALLENGE_DATA_MSG,
```
benliao1, line 43: this change will need to be communicated to all of the other teams, since this means that `LOG_MSG`, `DEVICE_DATA_MSG`, `GAME_STATE_MSG`, etc. are all shifted down by one number. if we don't want to do that / don't want to break backwards compatibility / want to deal with it later, we should leave `CHALLENGE_DATA_MSG` in this enum or replace it with something else, like `NOP_MSG` or something and mark it with a `TODO`

benliao1 (changes requested):
Pretty good! I left a few comments / questions that I wanted answers to, but the only thing that really needs changing is the deletion of `CHALLENGE_DATA_MSG` in `net_util.h`; the comment explains why.

tests/integration/tc_71_8.c
levincent06: Yep this was implemented as a misc change in #185!

net_handler/net_util.h
levincent06: I think this is a good time to communicate this with the other teams

net_handler/udp_conn.c
levincent06: This is also from PR #185

net_handler/udp_conn.c
levincent06: This is also from PR #185

net_handler/tcp_conn.c
AVSurfer123: I think I prefer how its written now.

benliao1 (approved):
Looks good, glad that this can finally be merged in!

AVSurfer123 (approved):
LGTM

PR #226, opened 2021-07-14
[MISC] Add protos as submodule

benliao1:
Added protos repo ([github.com/pioneers/protos](url)) as a submodule to runtime at `net_handler/protos`.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "net_handler/protos"]
```
DanielMolina24, line 1: Should we add a comment here about submodules or make it a section in the wiki?

DanielMolina24 (approved):

.gitmodules
benliao1: Yea I'll add a section describing submodules in the wiki, maybe on the Git page or something?

PR #234, opened 2021-09-17
[DEV_HANDLER][NET_HANDLER][SHM] Remove device subscriptions

### LINKED ISSUE #214: [EXECUTOR] Handle subscription edge case with Robot.get_value() in block comment
### Observed Behavior
If `studentcode.py` has the following code
```
def teleop_main():
      """
      Robot.get_value(<dev_id_here>, <param_name>, <val>)
      """
```
The subscription code parser won't realize that the API call is part of a comment. As a result, executor will try to subscribe to the device, which will throw an error if the device isn't connected.

### Expected behavior
Executor should recognize this as part of a comment, and ignore it when parsing for subscription requests.

### LINKED ISSUE #215: [EVERYTHING] Remove subscription maps from dev handler, executor, net handler protos
Subscription maps aren't very useful and only complicate the code base, so we want to remove them. This also means we can remove the executor's code parsing for subscriptions which didn't handle many edge cases. This will also close #214

benliao1:
Removes device subscriptions from Runtime. The feature was a huge complexity addition to Runtime and does not offer that much performance gain, so we simply removed it. Runtime will no longer be able to subscribe to a subset of the parameters for a device; it will always read all parameters of all devices and send data about all parameters of all devices to Dawn.

Closes #215 
Closes #214  (issue becomes obsolete)


net_handler/message.c
```diff
@@ -109,7 +108,6 @@ void send_device_data(int dawn_socket_fd, uint64_t dawn_start_time) {
 
     //get information
     get_catalog(&catalog);
-    get_sub_requests(sub_map, NET_HANDLER);
```
levincent06, line 112: We need to determine the bitmap passed into `device_read_uid()`. The bitmap should have bits on only for parameters that 1) exist and 2) are readable. An inefficient way to do this is to iterate through every single device and get this information from `runtime_util.h` functions.

levincent06 (changes requested):

levincent06:
Where does `studentapi.pyx` make sure that `Robot.get_value()` is valid only for readable parameters and `Robot.set_value()` is valid only for writeable parameters?

.gitignore
```diff
@@ -1,3 +1,18 @@
...
+
+# unignore all directories
+!*/
+
+# unignore all Makefile
+!*/Makefile
+
+# unignore all Dockerfile
+!*/Dockerfile
+
```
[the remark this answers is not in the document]
benliao1: @levincent06 Actually no, I forgot that we can't remove these because if we do that then it tries to add all of the executables as well. This is why we should probably put all of our executables into a separate `build` folder or something so that we can just ignore that folder and not worry about the other stuff. But that can be another issue.....

net_handler/message.c
benliao1: Update: we are discussing what to do about this on Slack; will post an update once this has been resolved.

levincent06: The solution we came up with is using a `get_readable_param_bitmap()` util function.

levincent06 (approved):
hasta la vista subscriptions

PR #237, opened 2021-09-22
[SYSTEMD][BUILD] New set_time.service and Travis builds cli

### LINKED ISSUE #220: [MISC][SYSTEMD] Update Raspberry Pi date and time on startup
During Spring Comp 2021, we've found that the clock on Raspberry Pis get unsynced when connected to CalVisitor+Motherbase. (The clock can be behind by days.) This becomes a problem for having accurate timestamps on logs and possible network issues that require synced times.

Below is a one-line command that will update the clock to the current time automatically. Perhaps we can add this to a systemd service so that this runs every time the Raspberry Pi boots up.
```
sudo date -s "$(curl -s --head http://google.com | grep ^Date: | sed 's/Date: //g')"
```

benliao1:
`set_time.service` and `set_time.sh` together will set the time on Raspberry Pi's because if they're connected to CalVisitor or Motherbase connected to CalVisitor, the raspberry pi's will not be able to access the NTP servers to get the current system clock, causing confusing Github messages and the like (certificate expired, certificate inactivated, etc.).

Travis used to not build the cli, which caused us to miss the dual `get_next_dev_data()` in `net_handler_cli` for months. Travis now builds the `cli` to make sure the they build properly in order to merge.

Closes #220

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,6 +10,7 @@ RUN echo 'export PATH=$PATH:~/runtime' >> ~/.bashrc
 
 # Build Runtime processes
 RUN ./runtime build
+RUN cd tests && make cli && cd .. # runtime build does not build the cli's; we want to build this in case something fails here
 
 # By default, run Runtime when the Docker container starts
 CMD ./runtime run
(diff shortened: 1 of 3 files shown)


levincent06 (approved):
accurate logs pog

benliao1:
Screen Shot 2021-09-22 at 4 42 10 PM

Here is proof that Travis is compiling the tests^^

PR #241, opened 2021-09-23
[DOCS] Removed docs folder because they're taking up too much space

benliao1:
In the process of removing large files from repo, we decided to remove the `docs` folder.

- Moved `style_guide.c` to wiki
- Uploaded pictures and graffles to a google drive folder


levincent06 (approved):
bye bytes

PR #243, opened 2021-09-26
[NET_HANDLER][TEST] Added WebSocket to TCP converter

AVSurfer123:
- Added new websocket_converter.py file to convert websocket comm
- Added websocket_client.py file to emulate a websocket client from Dawn
- Moved keyboard_interface.c functions into net_handler_cli
- Added pthread_cancel to stop_net_handler
- Added help message to net_handler_cli and automatically lowercase all inputs to the CLI
- When attaching to only 1 client, have the CLI automatically choose either Dawn or Shepherd instead of asking

Issues:
- [FIXED] Converter works only for first few commands sent (fixed by disabling ping interval heartbeat future)
- net_handler_client isn't receiving dev data from Runtime
- [FIXED] When sending start position in net_handler_cli, receiving `writen: issue sending start position message to dawn`


benliao1:
It's been so long since this has been worked on that I will close this pull request.. if in the future we want to use WebSockets to communicate with Dawn, we can dig out this PR again!

Issue #236, opened 2021-09-22
[DEV_HANDLER][LOWCAR] Add Fin message


benliao1:
When one side of the connection between the Raspberry Pi and the device times out, the side that timed out should send a FIN on the wire to the receiver to potentially let them know that they have been timed out and should therefore reset the connection. Currently, this is implemented as a timeout on both sides (i.e. when one side times out, it will wait for a predetermined amount of time to give the other end time to timeout and reset), but this is inconsistent, as we found in PR #234.

levincent06:
We closed this by accident

levincent06:
Dev Handler's exit handler should send a `FIN` to any currently connected devices. This should prevent connected motors from spazzing out whenever we `pkill -6 dev_handler`

PR #245, opened 2021-09-26
[LOWCAR][DEV_HANDLER] Remove LED test in KoalaBear

levincent06:
- Function call adds a sleep(1) that causes timeout issue
- Timeout reduced back to 1000ms, which was increased to 2000s as a short term solution

Issue #236 was originally thought to be the reason behind the timeout issue. This PR won't close that issue because the issue addresses the hacky solution of forcing a lowcar device to reset by pretending that dev handler dies.

### DIFF
--- a/lowcar/devices/KoalaBear/KoalaBear.cpp
+++ b/lowcar/devices/KoalaBear/KoalaBear.cpp
@@ -273,7 +273,6 @@ void KoalaBear::device_enable() {
     this->pid_enabled_b = TRUE;
 
     this->led->setup_LEDs();
-    this->led->test_LEDs();
 
     pinMode(AIN1, OUTPUT);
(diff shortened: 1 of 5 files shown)


lowcar/devices/KoalaBear/KoalaBear.cpp
```diff
@@ -273,7 +273,6 @@ void KoalaBear::device_enable() {
     this->pid_enabled_b = TRUE;
 
     this->led->setup_LEDs();
-    this->led->test_LEDs();
```
benliao1, line 276: F :endme:

benliao1 (approved):
Looks good lol, good job on finding this..... 💯 
I am so dumb for leaving this in last August haha

PR #247, opened 2021-09-29
[EXECUTOR][SHM] Auto-reset game states and stop motors

### LINKED ISSUE #212: [EXECUTOR] Stop motors and servos, but don't IDLE on source disconnect
When an input source disconnects, the mode should continue running but `executor` should force all of the motors and servos to stop moving. This is so that for momentary Keyboard disconnects (or something, we're not really sure what is causing the bug at the moment), the robot is not put into `IDLE` prematurely, requiring Shepherd to send the robot back to `TELEOP` using the Emergency Start button during a match.

### LINKED ISSUE #213: [EXECUTOR] Set all game states to INACTIVE on IDLE
When a robot gets put into IDLE, either by Dawn or Shepherd, it should automatically set all game states to `INACTIVE`. This is to prevent the robot from having leftover game state from a previous match if the game ends with the game state active.

levincent06:
## Game states:
Resolves #213
When the run mode is switched to `IDLE`, the game states are no longer relevant, so all game states should be set to `INACTIVE`. This is achieved by adding an `atexit()` handler in `gamestate_filter.c` when the game state thread is created. This works because the subprocess (`AUTO`/`TELEOP`) that spawns the game state thread in `gamestate_filter.c` is killed when switching to `IDLE`.
 
## Motors:
Resolves #212 
If it's `TELEOP` and neither `GAMEPAD` nor `KEYBOARD` is connected, a) the motors should be stopped, and b) writing to motor velocity should be void. This prevents issues where user input is disconnected and the motors continue running, which will cause physical damage.

- Update `shm_wrapper.c` to stop the motors when robot description is changed such that the above condition becomes true
  - Ensures robot is stopped immediately regardless of student code content
- Update `studentapi.pyx` to void writes to motor velocities while above condition is true
  - Ensures robot *stays* stopped even if student code tries to write to the motor velocity

The two following student code files were tested:
```
def teleop_setup():
    # Write to velocity

def teleop_main():
    pass
```
```
def teleop_setup():
    pass

def teleop_main():
    # Write to velocity
```


executor/studentapi.pyx
```diff
@@ -337,6 +337,14 @@ cdef class Robot:
         if param_idx == -1:
             raise DeviceError(f"Invalid device parameter {param_name} for device type {device.name.decode('utf-8')}({device_type})")
 
+        # EDGE CASE: If it's TELEOP but no UserInput is connected, motor controller velocities should remain 0
+        if (strcmp(device.name, "KoalaBear") == 0 \
```
AVSurfer123, line 341: Does this actually work? From what I know `strcmp` and any C strings are only compatible with `bytes` literals in Python, not `str`. You probably should convert `device.name` to a unicode string first and compare using Python instead of C.

runtime_util/runtime_util.c
```diff
@@ -272,6 +272,27 @@ int8_t get_param_idx(uint8_t dev_type, char* param_name) {
...
 }
 
+param_id_t* get_params_to_kill(uint8_t* num_devices_with_params_to_kill) {
+    // Change this array as necessary
+    /**
+     * Parameters that should set to (and remain) 0, 0.0, or False, depending on the type,
+     * when the Robot should be emergency stopped.
+     * Ex: Robot may be emergency stopped if no user input is connected during TELEOP.
+     */
+    param_id_t params_to_kill[] = {
```
AVSurfer123, line 282: Make static, or make it a global variable probably.

net_handler/net_handler.c
```diff
@@ -117,6 +118,9 @@ int main() {
     }
     shm_init();
 
+    // TODO: Net Handler is in charge of regulating game states.
+    // Net Handler may not have this responsibility in the future.
+    start_gamestate_handler_thread();
```
AVSurfer123, line 123: This looks a little ugly. Why have the `gamestate_filter.c` be in the `executor` folder if it is being imported into the `net_handler` only? Either start the thread from the `executor` or move the source file to the `net_handler` directory.

levincent06: It's also clunky because `studentapi.pyx` uses `filter_device_write_uid()` which is in `gamestate_filter.c` We're thinking of changing the implementation and bake the game state handler into `shm_wrapper.c`, but probably after this PR, since we want it done before this upcoming week.

I tried putting it in `executor`, but then it makes it awkward to test if the test file doesn't use a student code. (if there's no student code in a test case, then `executor` doesn't spawn)

runtime_util/runtime_util.c
levincent06: Is it possible to make it static if it uses functions? It used to be a global variable and I moved it into a function because I was getting errors/warnings

executor/studentapi.pyx
levincent06: It works -- is there an advantage to using unicode strings instead?

tests/client/net_handler_client.c
```diff
@@ -471,6 +471,61 @@ void send_user_input(uint64_t buttons, float joystick_vals[4], robot_desc_field_
...
+        printf("disconnect_user_input: Error when sending UserInput message");
+        exit(1);
+    }
+
+    // free everything
+    free(gamepad.axes);
+    free(inputs.inputs);
+    free(send_buf);
+}
+
```
benliao1, line 528: Is there a way to reconnect inputs once you disconnect them?

benliao1: Oh wait, I see it, `send_user_input` automatically connects the input. Maybe we should separate that out into its own separate function so that it's more obvious?

net_handler/net_handler.c
benliao1: I have an additional idea for this after much thinking (check Slack) that we can discuss?

tests/client/net_handler_client.c
levincent06: Do you mean `connect_user_input()` and `send_user_input()`, where the latter works only if the input is already connected? It will be more verbose and require updating a lot of the tests we already have

runtime_util/runtime_util.c
levincent06: I'm getting `initializer element is not constant` for using things like `KoalaBear.type` and `get_param_idx()` if I try to make the array static

tests/client/net_handler_client.c
benliao1: Yeah, I guess it's a little too verbose, we can assume that when we want to send user input, we connect the input.

PR #255, opened 2021-12-02
[LOWCAR] updated time for recieving device writes as well

### LINKED ISSUE #246: [LOWCAR] Update last_recv_msg_time on both DEVICE_PING and DEVICE_WRITE
We consider dev handler as "timing out" if a lowcar device hasn't received a message from dev handler in `TIMEOUT` milliseconds. Currently, we record the timestamp of the most recent `DEVICE_PING` to determine this. We should update this timestamp when we receive a `DEVICE_WRITE` as well.

This fix should need only a few lines of modifying `Device.cpp`

ElPescadoPerezoso1291:
RESOLVED #246
Added a line under Device Write case to update time every time we receive a device write ping.

### DIFF
--- a/lowcar/devices/Device/Device.cpp
+++ b/lowcar/devices/Device/Device.cpp
@@ -39,6 +39,7 @@ void Device::loop() {
                 break;
 
             case MessageID::DEVICE_WRITE:
+                this->last_received_ping_time = this->curr_time;
                 device_write_params(&(this->curr_msg));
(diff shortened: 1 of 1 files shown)


levincent06 (approved):
Looks good! Nice job on your first PR for Runtime 💯

benliao1:
And this one should also be good to merge.

PR #256, opened 2021-12-02
[LOWCAR] make sure lowcar parameter is valid

### LINKED ISSUE #253: [LOWCAR] Make sure param input is valid
We should make sure that `device_read()`/`device_write()` for all the devices return `0` if the `param` argument is invalid.

In `LimitSwitch::device_read()`, we call `digitalRead()` with the argument `param`, even if might not refer to a relevant pin.

wongquijote:
RESOLVED #253 
We should make sure that device_read()/device_write() for all the devices return 0 if the param argument is invalid.

In LimitSwitch::device_read(), we call digitalRead() with the argument param, even if might not refer to a relevant pin.


levincent06 (approved):
Looks good!

benliao1:
Should merge this one as well, this one is ready to go once the workflows CI is added and we rebase it on master

PR #258, opened 2022-01-06
[DEV_HANDLER][LOWCAR] Add RST message to Lowcar protocol

levincent06:
A RST message is a message with no payload, sent by either Dev Handler or Lowcar to indicate that the connection will be closed. This removes the jank code in dev handler that makes lowcar disconnect by sleeping (which makes lowcar thinks Dev handler is unresponsive).

For both Dev Handler and Lowcar:
- Before terminating Lowcar protocol, send a RST message.
- When a RST message is received, terminate the connection and clean up.

Resolves Issue #236


benliao1 (approved):
This looks good! Sorry I seem to have forgotten about this PR :\\ I suppose we need to test it to confirm that it works on an actual lowcar device but I know we test a lot of disconnections in our CI tests; if those pass, I'm confident that this code is solid.

PR #261, opened 2023-04-26
[DEV_HANDLER] [NETWORK_SWITCH] [LOWCAR] [SYSTEMD] Changes for 2023 Competition

jisungk2:
Created bin folder to move executables into.
Deleted BatteryBuzzer in lowcar/devices and replaced it with PDB. 
Added PDB to the lowcar/devices folder for battery display. 
Network switch is handled by the PDB switch and network_switch uses bash script to switch between local and pioneers router.
Updated dev_handler to recognize ttyUSB data from the PDB.
GeneralSerial class added in lowcar/devices/Device to process data using Serial1 and Serial.
Updated Makefiles and systemd files for network_switch, updating ./runtime commands and moving executables into bin.
Updated all paths and users from pi to ubuntu because of le potato in systemd.


lowcar/devices/Device/defs.h
```diff
@@ -58,7 +58,8 @@ enum class DeviceType : uint8_t {
     BATTERY_BUZZER = 0x03,
     SERVO_CONTROL = 0x04,
     POLAR_BEAR = 0x05,
```
benliao1, line 60: We added a new device called `PDB` here.

network_switch/Makefile
```diff
@@ -0,0 +1,11 @@
+LIBFLAGS=-pthread -lrt -Wall
```
benliao1, line 1: Here is the network switch code; all of the folders under `runtime/network_switch` were created on this branch.

network_switch/teamrouter.txt
```diff
@@ -0,0 +1,2 @@
+Team##
```
benliao1, line 1: Change this file for each team's router and password!

scripts/flash.sh
```diff
@@ -351,7 +351,7 @@ printf "\nAttempting to flash a $DEVICE_NAME!\n\n"
 arduino-cli config init
 
 # update the core index (see arduino-cli documentation)
-arduino-cli core update-index
```
benliao1, line 354: An update to the `arduino-cli` tool necessitated this change.

scripts/run.sh
```diff
@@ -5,20 +5,22 @@
...
-cd net_handler && ./net_handler &
+cd bin && ./net_handler &
 sleep 0.5
-cd executor && ./executor &
+cd bin && ./executor &
 sleep 0.5
-cd dev_handler && ./dev_handler &
+cd bin && ./dev_handler &
+sleep 0.5
+cd bin && ./network_switch &
```
benliao1, line 24: These changes guaranteed we clean the executables not in the `runtime/bin` directories from devices with older installations of Runtime.

systemd/dev_handler.service
```diff
@@ -1,15 +1,15 @@
 [Unit]
 Description=Device handler process for Runtime
-BindsTo=executor.service net_handler.service
+BindsTo=executor.service net_handler.service network_switch.service
 After=shm_start.service
```
benliao1, line 4: We had to modify all of the `systemd` service files to reflect the new home directory of the default user on the Le Potato, which was `ubuntu`. The home directory is now `/home/ubuntu` instead of `/home/pi`, as it was on the Raspberry Pi's.

benliao1 (approved):
As a general rule of thumb, this PR/branch is not a good example of software development best practices, since we had:
- multiple people pushing to the same branch
- multiple issues being worked on simultaneously on said branch
- changes to the build process throughout the branch

However, this was the first time that new members @jisungk2 @qwyl @BrandonWong14  worked on something substantial for Runtime, so it's ok! In the future we will follow software development best practices more closely. 

Great job team!! :D

PR #262, opened 2023-04-26
[LOWCAR] Small fixes to KoalaBear

jisungk2:
Removed device_enable from PDB.h and set pid_enabled to false for KoalaBear.cpp.


benliao1 (approved):
Looks good! Thanks for fixing~

PR #263, opened 2023-04-29
[BUILD] Replace Travis + Docker with Github Actions for CI

levincent06:
We have been using TravisCI in conjunction with Docker images in order to run our Runtime test script with every PR. Because we want to minimize our usage of third-party dependencies and keep things simple, we will stop using these two services in favor of Github Actions/Workflows.

A new .`yml` file defines the rules for CI. At a high level:
For every PR that wants to merge a branch into master:
1. Install all Runtime dependencies (protobufs, python, Cython).
2. Run a script to build Runtime
3. Run a script to test Runtime

If any step fails (ex: a test case fails), then the CI build will fail and block the PR from being merged.

Note that the setup of installing dependencies happens on the same VM that is reused for all CI builds, so the installation step will be relatively fast.


benliao1:
Am just testing some stuff over and over again....

benliao1:
We have one working test! The issue was line 172 of `dev_handler_client.c`, which uses `sprintf` to print the file descriptor for the virtual device into a buffer which is used as an argument before a `fork` and `exec` to spawn the new virtual device process. We didn't account for the fact that the file descriptor could be double-digits, so we made that the buffer only two bytes long, enough to fit a single-digit file descriptor and a null terminator. However, if the file descriptor is two digits, the buffer needs to be at least *3 bytes long* in order to fit the null terminator. I changed it to be 4 bytes to be 100% certain the null terminator would be there. 

Good example of a literal stack overflow error! (Took me 3.5 weeks to debug D: )

benliao1:
Cleaned up the branch a LOT and put accurate commit messages

benliao1:
Cleaned up branch some more, added scheduled trigger of CI every 7 days so that we don't lose the cache, and updated `README.md`. For some reason the checking if the `cache-hit == 'false'` doesn't work; we have to use `cache-hit != 'true'` which is really dumb...

benliao1:
I attempted to shave off some time by caching the dev package files as well, but it doesn't save much time at all, if anything. The dev package files are like 400 MB, which take about 15 seconds to load into the runner from the cache anyways; it's about the same time to just run the `sudo apt-get` commands anyways. So I decided not to try to cache them for simplicity.

I did, however, manage to remove some things such as `python3-pip` from our install list (they are pre-installed on the Github runner)

jisungk2 (approved):
perfect

PR #265, opened 2023-06-18
[BUILD] Removed cron trigger for workflow

benliao1:
The idea behind putting a schedule on the CI build was to make sure that the cache never expired. However, since it seems that a cache is only valid when run on the same PR/same branch, this doesn’t matter since we won’t be able to transfer the cache between PRs or between a working branch and master. I removed it in this PR.


BrandonWong14 (approved):
perfect 2.0

Issue #229, opened 2021-09-16
[NET_HANDLER] Remove MODE__ESTOP


benliao1:
The run mode proto still has the ESTOP run mode, which is deprecated. The change should be made in the `protos` repo and announced to the other software teams for adoption.

benliao1:
Closed by #271 . Closing now.

PR #271, opened 2023-08-07
[MISC] Cleanup

benliao1:
This PR simply deletes some old unused code and makes some changes to the `.gitattributes` file so that the Language statistics on our home page more accurately reflect the contents of our repo. Specifically, this PR does the following:

- Fixes a bug in the `.gitattributes` file which previously still included the files in `lowcar/lib` as Runtime code, and which previously classified some files in `lowcar/devices` as C code, when they are C++. It also specifically tells Github to highlight `Makefile_defs` with Makefile syntax
- Bumps the `net_handler/protos` submodule to the newest version in that repo, including updates to the `run_mode.proto` and `security.proto`
- Resolves issue #229 remove the run mode `ESTOP` from Runtime
- Fixes the formatting in `scripts/test.sh`

In `lowcar`, the RFID and PolarBear devices were deleted, as well as the MFRC22, SevenSeg, and SPI libraries. If new libraries need to be added, they can be downloaded from the Arduino forums and added back into the `lowcar` library.


BrandonWong14 (approved):
Seems good makes sense, the number of files being removed seemed intimidating until I realized most were different parts of very few individual things lol

PR #274, opened 2024-03-07
[SYSTEMD] updated last few files from pi to ubuntu

BrandonWong14:
Updated some systemd files that still had pi in the path and username instead of ubuntu. It's probably not completely necessary, but it is good that it is fully updated. I am also using this as an example of a pull request to show new members.


3142008956 (approved):
Everything looks good.

BrandonWong14:
Modified the yml file in .github/workflows/, moved the install protobuf-c check after the format check so format does not look at protobuf-c files as well. checkout and cache have been changed to v4 as well in the file.

PR #276, opened 2024-05-28
[EXECUTOR] Updated Python version to 3.11

BrandonWong14:
I just updated the Python version in executor so we don't have to manually change it on the Raspberry Pi's like we had to do this past spring.


BrandonWong14:
Please ignore this, figuring things out.

BrandonWong14:
more changes have been made since including this, this is no longer necessary

PR #277, opened 2024-06-12
Updated last 4 student code test files 

3142008956:
Updated last 4 files in tests/student_code to account for the updated executor. Replaced autonomous setup/main with autonomous() and teleop setup/main with teleop(), and added while True loops to simulate continuous execution.


BrandonWong14 (approved):
Looks good, we can deal with the failing tests to see if its still failing after the rest are updated.

PR #282, opened 2024-07-06
(Ishan) Redid the first 5 student tests

BrandonWong14:


BrandonWong14:
executor_rewrite is chaotic enough right now, these changes are simple enough that they can be retackled later if necessary

PR #285, opened 2025-03-14
[EXECUTOR] Modified executor to place loops on student end

BrandonWong14:
The primary update in this pull request is updating executor such that loops now must be written by the students rather than done in executor itself. This was primarily a simplification of the code, ideally to deal with rare bugs around sleeping. It should also help students understand how exactly their code is running better.
Additionally, python version has been set to 3.11, the rasbian default. Tests have been updated to account for the new executor and python version.


3142008956 (approved):
All tests pass, if anything else comes up, we all blame Brandon.

ishan-monie (approved):

